PR #46, opened 2022-05-18
Umls mesh

cbizon:
In reviewing some disease/phenotype clusters, I discovered some over-merging caused by overagressive UMLS/MESH relationships.  It's not uncommon to have multiple UMLS annotated to a single MESH.  This led to some incorrect merges.  So for instance, MESH:D015835  has multiple UMLS annotations in MRCONSO.RRF.  We ended up associating this mesh term, which is equivalent to a particular MONDO to a related but not equivalent UMLS.  

This PR cleans this up by asserting that UMLS must be mapped to the Main Heading in mesh, not just to some synonym.

In the process, I discovered that changes to MONDO or Ubergraph had broken OMIM and ICD10 identifiers, so I fixed those as well.


gaurav (approved):
LGTM! I was hoping to finish a run of Babel with this new code to make sure that it works entirely and so I can compare its output to the previous run, but it's still running on Sterling at the moment.

cbizon:
You may find some issues with disease / phenotype b/c OMIM and ICD10 mappings from MONDO had changed format, but this change should fix it.

PR #49, opened 2022-07-18
Paginate UberGraph results and remove EFO

gaurav:
Large UberGraph queries continue to time out. This PR adds code to paginate UberGraph queries so that we can retrieve larger result sets.

It also removes EFO from the list of UberGraph ontologies as removed in https://github.com/INCATools/ubergraph/pull/66.

### DIFF
--- a/config.json
+++ b/config.json
@@ -26,9 +26,9 @@
 
   "geneprotein_output": "GeneProtein.txt",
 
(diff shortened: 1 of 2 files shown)


gaurav:
We're now tracking putting EFO back in at #58.

PR #50, opened 2022-07-18
Improve UMLS coverage

gaurav:
This PR will implement the plan in https://github.com/TranslatorSRI/NodeNormalization/issues/119#issuecomment-1154751451 to (1) create a compendium of UMLS IDs, labels and types for UMLS IDs that weren't included in other compendia, and (2) produce a report that will help us find sections of the UMLS that should be incorporated into Babel in other compendia.

It also fixes a minor bug in which "UMLS" was not listed as a prerequisite for "chemical", which caused "chemical" in some cases to run before the UMLS ids, labels and synonyms had been produced.


src/createcompendia/leftover_umls.py
```diff
@@ -0,0 +1,239 @@
...
+
+                # How to deal with multiple Biolink types? We currently only have the following multiple
+                # types, so we can resolve these manually:
+                biolink_types_as_str = '|'.join(sorted(map(lambda t: "(None)" if t is None else t, list(biolink_types))))
+                if None in biolink_types:
+                    # One of the TUIs couldn't be converted; let's delete all of them so that we can report this.
+                    biolink_types = list()
+
+                # Some Biolink multiple types we handle manually.
+                if biolink_types_as_str == 'biolink:Device|biolink:Drug':
```
cbizon, line 150: I think it would be good to use the constants in src/categories.py.   The goal is to insulate against name changes, though to be fair I don't know that's a big problem w/biolink.

src/createcompendia/leftover_umls.py
```diff
@@ -0,0 +1,239 @@
...
+                umls_id = f"{UMLS}:{cui}"
+                if umls_id in umls_ids_in_other_compendia:
+                    logging.debug(f"UMLS ID {umls_id} is in another compendium, skipping.")
+                    continue
+                if umls_id in umls_ids_in_this_compendium:
+                    logging.debug(f"UMLS ID {umls_id} has already been included in this compendium, skipping.")
+                    continue
+                lang = x[1]
+
+                # TODO: note that this technically skips terms that don't have any English labels.
```
cbizon, line 114: Is there a way to have this suppression/language logic in one place?   We have changed it a few times, so I'm worried about it getting fixed in one place and not another.

cbizon (changes requested):

src/createcompendia/leftover_umls.py
gaurav: Fixed in 6db455aee8b2adce.

src/createcompendia/leftover_umls.py
gaurav: Centralized the suppression/language code in 335c010b733361954.

cbizon (approved):

PR #53, opened 2022-08-03
Add complex portal

### LINKED ISSUE #52: Add complex portal to downloads
* Add a complex portal section to src/snakefiles/datacollect.snakefile
* Create a complex portal ingester to do the work in src.datahandlers to download, extract labels and synonyms
* Decide on the biolink class (Cellular Component or Macromolecular Complex)
* Add the appropriate prefix to biolink

Shalsh23:
Fixes #52


Shalsh23:
- [x] Pull yeast complexes form complex portal
- [ ] Parse yeast complexes for labels and synonyms

src/prefixes.py
```diff
@@ -69,3 +69,4 @@
 NCBITAXON='NCBITaxon'
 HGNCFAMILY='HGNC.FAMILY'
 PANTHERFAMILY='PANTHER.FAMILY'
+COMPLEXPORTAL='ComplexPortal'
```
Shalsh23, line 72: @cbizon Should I update this to the biolink prefix `MacroMolecular Entity`?

Shalsh23:
@cbizon @jdr0887 I'll appreciate any feedback you might have on the pythonic way of writing code and other coding best practices as well.

src/datahandlers/complexportal.py
```diff
@@ -0,0 +1,31 @@
...
+    infile = open(infile, 'r')
+    lines = infile.readlines()[1:]
+    usedsyns = set()
+
+    with open(labelfile, 'w') as outl, open(synfile, 'w') as outsyn:
+        for line in lines:
+            sline = line.split("\t")
+            id = sline[0]
+            label = sline[1] # recommended name
+            outl.write(f'Complex ac:{id}\t{label}\n')
```
Shalsh23, line 17: This needs to be updated per the biolink prefix.

src/prefixes.py
cbizon: I'm not sure I understand.  Macromolecular entity is not a prefix, it's a type.  What we need to do is add ComplexPortal as a prefix to the Macromolecular Machine type in biolink.

cbizon (commented):
This is basically fine.  One minor change I would probalby make would be to add the input file to the `with` context for the output files and then use `for line in infile`.   That way you have a much more memory efficient code because you'll only be storing one line at a time.

src/datahandlers/complexportal.py
cbizon: have we requested its addition?

cbizon: Either way, we should replace it here with the constant from prefixes.py

jdr0887 (approved):

PR #58, opened 2022-08-21
Add EFO

### LINKED ISSUE #48: Handle missing efo
Jim has removed EFO from ubergraph, we will need to get it directly rather than from ubergraph

gaurav:
WIP: This PR is an attempt to fix #48 by importing EFO directly into Babel. The challenge is that EFO "combines parts of several biological ontologies, such as UBERON anatomy, ChEBI chemical compounds, and Cell Ontology", all of which we pull in from other sources (mainly Ubergraph). So rather than pulling everything from EFO as we might with another ontology, we should probably try to selectively choose the parts of this ontology that Babel needs and that isn't already covered by another ontology.

### DIFF
--- a/src/createcompendia/diseasephenotype.py
+++ b/src/createcompendia/diseasephenotype.py
@@ -124,7 +124,7 @@ def build_disease_umls_relationships(idfile,outfile,omimfile,ncitfile):
             for line in inf:
                 x = line.split()[0]
                 good_ids[prefix].add(x)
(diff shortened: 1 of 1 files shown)


matentzn:
One thing you could do is use ROBOT to create a base, i.e. remove all content that does not relate to terms in the EFO namespace. Pulling in information about terms _other than_ terms in the EFO namespace is probably not a great idea as many of the terms have not been updated in 10 years. I would also take EFO xrefs with a grain of salt, in case you need them for babel. If you want a suitable robot command to prune EFO according to what I think is right for your use case, let me know.

gaurav:
This approach to re-including EFO is probably not going to work (see #48). Closing.

PR #64, opened 2022-09-14
Improve kegg requests

jdr0887:


src/datahandlers/kegg.py
```diff
@@ -18,111 +20,152 @@
...
-                        if name.endswith(';'):
+    try:
+
+        request_session = requests.Session()
+        request_adapter = requests.adapters.HTTPAdapter(max_retries=10)
+        request_session.mount("http://", request_adapter)
+        request_session.mount("https://", request_adapter)
+
+        # 1:22530 + 1 to have inclusive end
+        compound_chunks = list(chunked(range(1, 22531), 10))
```
gaurav, line 33: I think this is a good opportunity to document why we specifically get CPDs from 1:22530 here 😄 . Do you know the answer to that? If not, we can ask @cbizon.

gaurav (approved):
This looks great, and I appreciate the Black code formatting -- I've added an issue to eventually apply Black to this entire repo (https://github.com/TranslatorSRI/Babel/issues/65)! I've made one minor comment regarding the provenance of the `1:22530` CPD IDs at the heart of this function -- once you've made those changes, feel free to merge this PR in!

src/datahandlers/kegg.py
jdr0887: I don't know the answer to this, but I can speculate.  I believe it's that the entries are added & the collection grows.  Previous commit was a range from 1:22250 & when I was looking the kegg api, no more data was available after 22530...indicating that this a moving target.  I could not find an endpoint that specified the current range such that we don't have to hard code it.

cbizon: Unfortunately, there's no way to know the right answer here.  There's not an API or anything that lets you know this number.  So that's an experimentally determined value, and potentially a moving target.   KEGG used to make data tables available for download, but stopped that practice a number of years ago.

cbizon (approved):

PR #113, opened 2023-03-10
change syn file format

cbizon:
Changes the structure of the synonyms files to 
1. Be in jsonl format 
2. include a preferred_name label
3. include biolink types

This new style can be loaded into the prototype name resolver.

While working on this PR, we also fixed a bug in NCBIGene synonym generation.

### DIFF
--- a/src/babel_utils.py
+++ b/src/babel_utils.py
@@ -235,8 +235,8 @@ def write_compendium(synonym_list,ofname,node_type,labels={},extra_prefixes=[]):
                     if "label" in node["identifiers"][0]:
                         document["preferred_name"] = node["identifiers"][0]["label"]
                     sfile.write( document )
-                except Exception:
-                    print('Hey')
+                except Exception as ex:
+                    print(f"Exception thrown while write_compendium() was generating {ofname}: {ex}")
                     print(node["type"])
                     print(node_factory.get_ancestors(node["type"]))
                     exit()
--- a/src/babel_utils.py
+++ b/src/babel_utils.py
@@ -230,9 +230,10 @@ def write_compendium(synonym_list,ofname,node_type,labels={},extra_prefixes=[]):
                 synonyms_list = sorted(synonyms,key=lambda x:len(x))
                 try:
                     document = {"curie": node["identifiers"][0]["identifier"],
-                                "preferred_name": node["identifiers"][0]["label"],
                                 "names": synonyms_list,
                                 "types": [ t[8:] for t in node_factory.get_ancestors(node["type"])]} #remove biolink:
+                    if "label" in node["identifiers"][0]:
+                        document["preferred_name"] = node["identifiers"][0]["label"]
                     sfile.write( document )
                 except Exception:
                     print('Hey')
--- a/src/babel_utils.py
+++ b/src/babel_utils.py
@@ -226,7 +226,10 @@ def write_compendium(synonym_list,ofname,node_type,labels={},extra_prefixes=[]):
                     nw['ic'] = ic
                 nw['identifiers'] = [ {k[0]:v for k,v in nids.items()} for nids in node['identifiers']]
                 outf.write( nw )
-                synonyms = synonym_factory.get_synonyms(node)
+
+                # get_synonyms() returns tuples in the form ('http://www.geneontology.org/formats/oboInOwl#hasExactSynonym', 'Caudal articular process of eighteenth thoracic vertebra')
+                # But we're only interested in the synonyms themselves, so we can skip the relationship for now.
+                synonyms = [result[1] for result in synonym_factory.get_synonyms(node)]
                 synonyms_list = sorted(synonyms,key=lambda x:len(x))
                 try:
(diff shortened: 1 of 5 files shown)


src/babel_utils.py
```diff
@@ -227,9 +227,12 @@ def write_compendium(synonym_list,ofname,node_type,labels={},extra_prefixes=[]):
                 nw['identifiers'] = [ {k[0]:v for k,v in nids.items()} for nids in node['identifiers']]
                 outf.write( nw )
                 synonyms = synonym_factory.get_synonyms(node)
-                if len(synonyms) > 0:
-                    for p,o in synonyms:
-                        sfile.write(f'{node["identifiers"][0]["identifier"]}\t{p}\t{o}\n')
+                synonyms.sort(key=lambda x:len(x))
```
gaurav, line 230: `synonyms` is a set(), so it doesn't have a `.sort()` method (giving the error: `'set' object has no attribute 'sort'`). I think you'll need to use something like https://stackoverflow.com/a/58820959/27310 to sort it.

src/babel_utils.py
```diff
@@ -227,9 +227,12 @@ def write_compendium(synonym_list,ofname,node_type,labels={},extra_prefixes=[]):
...
+                document = {"curie": node["identifiers"][0]["identifier"],
+                            "preferred_name": node["identifiers"][0]["label"],
+                            "name": synonyms,
```
gaurav, line 233: If this is going to be a list of synonyms and/or labels, I think the property name should be `names` rather than `name`.

gaurav (changes requested):
Looks good! This code doesn't currently run because synonyms is a set() and not a list() (see inline comment), but that should be a simple fix. I would also recommend changing `name` to `names` if it's going to be a set or list of names instead of a single name.

cbizon:
Can't fully test b/c of BMT bug, but I think that this fixes the problems.

gaurav:
Thanks, Chris! I'll wait for the BMT bug (https://github.com/biolink/biolink-model-toolkit/issues/111) to be fixed before reviewing this.

gaurav:
I tested this out on "anatomy" and it works great! Let me know if you'd like to have a look at the generated synonym files. Otherwise, I'm planning to rerun all of Babel early next week and can set up a NameRes loaded from the updated synonym files then.

gaurav:
One last thing that doesn't make sense: apparently this code (and my code) all worked fine even on bmt v0.9.0 today, not v1.0.4 as fixed by Sierra! I do not understand and do not like pip. I'll upgrade to v1.0.4 and recheck before approving this PR.

gaurav (approved):
I've confirmed that this works fine on bmt 1.0.4. I think we're good to merge!

gaurav:
I tried loading this into Solr, but I found something unexpected: note that `synonym_factory.get_synonyms(node)` doesn't return a list of synonyms, but a list of tuples in the form ('http://www.geneontology.org/formats/oboInOwl#hasExactSynonym', 'Caudal articular process of eighteenth thoracic vertebra'), where the first element is either oboInOwl:hasExactSynonym or the relation included in the synonyms file. When saved into JSON, this is turned into a two dimensional array:

```json
{ "names": [
  ["http://www.geneontology.org/formats/oboInOwl#hasExactSynonym", "Caudal articular process of eighteenth thoracic vertebra"],
  ["http://www.geneontology.org/formats/oboInOwl#hasExactSynonym", "Caudal articular process of eighteenth thoracic vertebra (body structure)]"]
] }
```

I assume we want to keep this format in Solr. However, when I try loading these JSON files into a Solr instance, it flattens this into a single list of strings:

```json
{"names":[
  "[http://www.geneontology.org/formats/oboInOwl#hasExactSynonym, Caudal articular process of eighteenth thoracic vertebra]",
  "[http://www.geneontology.org/formats/oboInOwl#hasExactSynonym, Caudal articular process of eighteenth thoracic vertebra (body structure)]"]
}
```

This still works, since you can still search for "Caudal" and get back this result, but all the names also match "hasExactSynonym", which seems non-ideal.

Do we want to keep the synonym types in the synonym output? If not, I could remove that we could simplify this to just a list of names. If we do, I can look into whether we can include more complex data into Solr. @cbizon What do you think?

gaurav:
- [ ] There are fewer synonyms being emitted for some terms (e.g. UBERON:0000178 has 13 synonyms in 2022dec2-2 but only five with this code). I wonder if we previously considered labels to also be synonyms but don't do that any longer, or if there's some other general reason for this drop.
- [ ] Some leftover UMLS synonyms produced with this code only had a single synonym (e.g. UMLS:C0000715 only has "House, Slaughter") but in 2022dec2-2 these had many more synonyms (e.g. UMLS:C0000715 had ten, including various spellings of "abattoirs" and "slaughterhouses"). This is probably a bug in the UMLS export I wrote.

gaurav (approved):
I'm pretty sure I've found and fixed all the immediate issues with this PR, so I think it's fine to merge it in. Some additional fixes are in PR #124.

PR #120, opened 2023-03-28
Returned EFO identifiers to disease/phenotypes

cbizon:


cbizon:
This PR returns EFOs to the disease/phenotype cliques.   It pulls efo.owl and uses it to extract labels/synonyms and exact equivalences.

It also does a little bit towards better prefix handling.   If there is inconsistent prefxing (like ORPHANET and orphanet) then it can lead to errors in merging.   This tries to make sure at curification time that the correct prefix is being used.  It doesn't go as far as it needs to (actually reading the biolink context file) but it helps.

src/datahandlers/efo.py
```diff
@@ -0,0 +1,116 @@
+from src.prefixes import EFO,ORPHANET
+from src.babel_utils import pull_via_urllib
+from src.babel_utils import make_local_name
+from src.util import Text
+import pyoxigraph
+
+
+
+def pull_efo():
+    _=pull_via_urllib('http://www.ebi.ac.uk/efo/','efo.owl', subpath='EFO', decompress=False)
```
gaurav, line 10: This is fascinating. What's the benefit of assigning it to underscore?

cbizon: Ah, I have seen that used in python specifically to denote that there is a value returned, but you're not going to do anything with it.  I don't use it consistently, for better or worse.

cbizon: e.g. https://towardsdatascience.com/5-different-meanings-of-underscore-in-python-3fafa6cd0379.

gaurav (approved):
Sorry it took me so long to get back around to this -- it looks good, and seems to work fine! Thank you!

PR #159, opened 2023-06-27
Adds a `shortest_name_length` value to the synonyms file

gaurav:
This restores the previously used sorting method as per https://github.com/TranslatorSRI/NameResolution/issues/65#issuecomment-1610056843.

WIP


gaurav:
This code has now been moved into PR #166, and so is not needed here. Closing.

Issue #125, opened 2023-04-19
Gzip compendium and synonym files


gaurav:
Compendia and synonym files take up a lot of storage and transfer time. It would be great if we could Gzip all the files as we write them out. One part of this is easy to do, since both files are created by [write_compendium()](https://github.com/TranslatorSRI/Babel/blob/782e77306e52c4947c6182302f99438db9405315/src/babel_utils.py#L202), so we can change that to produce Gzip pretty easily. However, to do this correctly, we should change the files to have `.jsonl.gz`, which would require changing the files in the individual Snakemakes and in all the `build_compendia()` functions.

PR #162, opened 2023-06-29
Increased storage on Babel

gaurav:
This PR increases storage on Kubernetes. This is probably not strictly necessary, especially if we could Gzip all output files (#125). But for now this is the quickest way forward.


gaurav:
This change will be made in PR #166, so this PR is not necessary. Closing.

PR #165, opened 2023-07-07
Added a test to warn on short names

gaurav:
This PR adds a warning for short names in the list of synonyms for a CURIE, as this could be a sign of incorrect synonyms being included in Babel. Also: we sometimes get duplicate synonyms, which I've just realized is because `synonym_factory.get_synonyms()` only ensures that `(relation, synonym)` tuples are unique. I've added a `set()` to ensure that the actual synonyms are unique, too.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,4 +1,6 @@
 {
+  "warn_on_short_names": 3,
+
   "input_directory": "input_data",
   "download_directory": "babel_downloads",
   "intermediate_directory": "babel_outputs/intermediate",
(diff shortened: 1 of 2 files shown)


config.json
```diff
@@ -1,4 +1,6 @@
 {
+  "warn_on_short_names": 3,
```
cbizon, line 2: how often does this happen?  I would assume we're going to see zillions of warnings...

gaurav: 769,715 times in 2023jul9, and 1,857,487 times in the current Babel... so yeah, probably too often to be useful :(. I'll close this PR and make a new one just to fix the [loophole in synonym uniquification](https://github.com/TranslatorSRI/Babel/pull/165/commits/54cfcdcecac3d5a60d7aedaa442becead40d06b4).

PR #134, opened 2023-05-14
Babel 1.0

gaurav:
Arbitrarily designates a version of Babel as Babel 1.0. This PR is intended to be useful during development; once all the other Babel PRs have been reviewed and merged, this should be unnecessary.

Should be merged after PR #97, PR #121, PR #132, PR #133, PR #135, PR #139, PR #165 and PR #166.

PR #168, opened 2023-07-11
Export information content in synonym files

gaurav:
As per https://github.com/TranslatorSRI/NameResolution/issues/72#issue-1782935627, we should try to see if we can use the information content values for cliques to improve sorting. This PR adds the information content value to the synonym export.

Should be merged after PR #134.


gaurav:
This is too far out of date to easily update to the latest version of Babel, so I'm going to close this. We do seem to be doing a good job with the searches now without considering information content values, but if this becomes necessary we should be able to reimplement this again without too much effort.

PR #173, opened 2023-07-25
Drugs

### LINKED ISSUE #144: Add RxNorm identifiers and metadata
- Ingredient RxNorm (but not other RxNorm ID types) map very close to UNIIs, so those mappings might be useful to incorporate, especially if it helps solving the "ibuprofen" vs "ibuprofen, tablet" vs "ibuprofen 5mg" problem.
- RxNorm metadata may also be useful for NodeNorm.

Asked by @jeffhhk

### LINKED ISSUE #150: Add RxNorm
RxNorm is a definitive source for drug information, and so we need to include RXCUIs in cliques.

The general algorithm we should use is:
1. Include all RxNorm entries, e.g. https://mor.nlm.nih.gov/RxNav/search?searchBy=RXCUI&searchTerm=204442
2. Consider them identical to all Ingredient (IN) identifiers -- in this case, [RXCUI 5640 "ibuprofen"](https://mor.nlm.nih.gov/RxNav/search?searchBy=RXCUI&searchTerm=5640). But note that this does not work with MINs, which work the other way, and tell you which drugs include this one.
3. Add additional names from RxNorm -- trade names, formulations, etc -- as synonyms of this concept.
4. Look for additional information in UMLS (e.g. https://uts.nlm.nih.gov/uts/umls/vocabulary/RXNORM/5640) such as trade names, formulations, etc.

This will have the side-benefit of getting rid of a lot of e.g. "Ibuprofen 50MG" entries in NameRes, since they would be cliqued into their ingredients. We could also extract RxNorm-UMLS and RxNorm-MeSH mappings (maybe to other sources as well: ChEBI? UNII?) so that we clique them together with other drugs.

Would close https://github.com/TranslatorSRI/NodeNormalization/issues/127

cbizon:
Fixes #150 
Fixes #144


cbizon:
This does basically 2 things - 
1. It adds in RXNORM identifiers and creates a Drug compendium
2. It pulls in RXNORM relationships and PUBCHEM RXNORM annotations and creates a drug/chemical conflation

PR #180, opened 2023-08-22
Drugs conflated synonyms

gaurav:
This PR adds some code for generating a "drug conflated synonyms" file, which includes all the synonyms from the chemical outputs except for those that have been conflated by the Drug conflator (PR #173).

WIP: I'm still testing this on Sterling, but I think it's worth reviewing the code to spot any obvious bugs or ideas for improving the algorithm.


src/synonyms/synonymconflation.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+
+    logging.info(f"conflate_synonyms({synonym_files}, {conflation_file}, {output})")
+
+    # Some common code to manage the conflation index.
+    # This is simply a large dictionary, where every key is an identifier and the value is the identifier to map it to.
+    conflation_index = dict()
+    conflations = defaultdict(list)
+
+    def add_conflation(primary_id, secondary_id):
```
cbizon, line 40: It's legal but functions defined in functions always freaks me out.

cbizon (approved):

src/synonyms/synonymconflation.py
gaurav: For bonus confusion, this function is only used in one place :). Removed in d15b79b.

PR #181, opened 2023-08-25
Improve conflation

cbizon:
I have identified 2 problems:

1. When we build the RXNORM conflation relations, we have some checks to make sure that we're not e.g. merging acetaminophen and caffeine because they are both ingredients in excedrin.  This takes the form of checking that a particular list has only a single element.  But it turns out that RXNREL can contain essentially the same triple multiple times. So that list should really be a set.   This PR fixes this issue.

2. There are some UMLS entities like "wellbutrin SR" that don't merge with anything in RXNORM or anywhere else.  My original hope was that we could use the UMLS REL file to connect these terms to other relevant terms.  However, the problem is that the only relatinons are "has_ingredient" and "has_tradename" relations that are subject to the rules mentioned above.  Because of the structure, these aren't able to merge in (there are MxN type has_ingredients edges, relaxing the constraints leads to everything is everything).  I'm not sure quite what to do with these.   I'm tempted to say that maybe NR should just not include e.g. drugs or chemicals that are a single UMLS term not merged with anything, but that's probably too draconian.

### DIFF
--- a/src/createcompendia/drugchemical.py
+++ b/src/createcompendia/drugchemical.py
@@ -224,7 +224,7 @@ def build_pubchem_relationships(infile,outfile):
             for cid in cids:
                 outf.write(f"{RXCUI}:{rxnid}\tlinked\t{PUBCHEMCOMPOUND}:{cid}\n")
 
-def build_conflation(rxn_concord,pubchem_rxn_concord,drug_compendium,chemical_compendia,outfilename):
+def build_conflation(rxn_concord,umls_concord,pubchem_rxn_concord,drug_compendium,chemical_compendia,outfilename):
     """RXN_concord contains relationshps between rxcuis that can be used to conflate
     Now we don't want all of them.  We want the ones that are between drugs and chemicals,
     and the ones between drugs and drugs.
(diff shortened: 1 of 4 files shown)


src/createcompendia/drugchemical.py
```diff
@@ -112,7 +112,11 @@ def get_cui(x,indicator_column,cui_column,aui_column,aui_to_cui,sdui_to_cui):
         if x[indicator_column] == "CUI":
             return x[cui_column]
         elif x[indicator_column] == "AUI":
-            return aui_to_cui[x[aui_column]]
+            try:
+                return aui_to_cui[x[aui_column]]
+            except:
+                #this really shouldn't happen.  But it seems to occur for the UMLS files?
```
gaurav, line 118: Would it be worth logging this so we can figure out what's going wrong later on?

gaurav (approved):
This runs successfully on Sterling, and -- although I don't really understand the code -- I don't see anything obviously wrong or odd. The output file hasn't changed much AFAICT: you can see the differences by comparing `/projects/babel/babel-outputs/2023aug3-with-drugs/conflation/DrugChemical.txt` (generated using this code) with `/projects/babel/babel-outputs/2023aug3-with-drugs/conflation/DrugChemical-2023aug18-before-pr-181.txt` (generated using the previous conflation code).

Also left a minor comment for your consideration.

gaurav:
> The output file hasn't changed much AFAICT: you can see the differences by comparing `/projects/babel/babel-outputs/2023aug3-with-drugs/conflation/DrugChemical.txt` (generated using this code) with `/projects/babel/babel-outputs/2023aug3-with-drugs/conflation/DrugChemical-2023aug18-before-pr-181.txt` (generated using the previous conflation code).

Ah no, I take that back -- I reran Babel from scratch and got a whole lot more changes. You can compare the new conflation file at `/projects/babel/babel-outputs/2023aug31/conflation/DrugChemical.txt` with the previous one at `/projects/babel/babel-outputs/2023aug3-with-drugs/conflation/DrugChemical-2023aug18-before-pr-181.txt` on Hatteras. I started a new Solr database load last night, and I expect it to finish sometime tonight. Once that's done, I'll set up a NameRes to point to it so we can look at the differences.

gaurav:
This is now up at https://nodenormalization-sri.renci.org/docs, but there appears to be a bug to do with which preferred ID we select when doing drug conflation (https://github.com/TranslatorSRI/NodeNormalization/issues/220). I'm going to go ahead and merge this because I don't think it's directly related to that bug, which can be dealt with separately.

PR #202, opened 2023-11-03
Improve UniProtKB downloads

gaurav:
UniProtKB downloads have gotten really slow lately. Rather than relying on the built-in `pull_via_urllib()` method, this PR switches that over to using `wget --continue` so that we get progress updates and resume incomplete downloads. This is a pretty hacky way of implementing this, but it's quicker than the two alternatives:
1. Rewriting `pull_via_urllib()` so that it can resume an incomplete download.
2. Writing a `pull_via_wget()` method that can either be a plug-in replacement for `pull_via_urllib()` or can be inserted into Snakemake.

Breaking up the downloads in Snakemake has the advantage that the files can be downloaded simultaneously.

WIP: after writing all this down, I think I should write a `pull_via_wget()` method to clean this up a bit before merging it.

### DIFF
--- a/src/datahandlers/uniprotkb.py
+++ b/src/datahandlers/uniprotkb.py
@@ -1,7 +1,5 @@
 from src.babel_utils import pull_via_urllib, make_local_name
 
-def pull_one_uniprotkb(which):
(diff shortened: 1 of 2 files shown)


cbizon (approved):
I'm surprised that there isn't a wget module for python, but as far as I can tell there ain't.

Issue #126, opened 2023-04-23
Replace leftover UMLS custom output generation with write_compendia()


gaurav:
The leftover UMLS file currently replicates the compendium and synonym generation code in babel_utils.write_compendia():

https://github.com/TranslatorSRI/Babel/blob/41d62bed0c009c0a62e45b91b209e854cee0831a/src/createcompendia/leftover_umls.py#L165-L195

This means that when the output formats change, we need to modify it in both write_compendia() and leftover UMLS.

We should replace this by having leftover UMLS generate an ID file (in e.g. `intermediate/leftover_umls/ids/UMLS`), and then use babel_uris.write_compendia() to generate the compendia and synonyms files.

PR #203, opened 2023-11-03
Skip writing synonyms that have no valid name

gaurav:
I'd been previously been keeping identifiers in the synonyms file that have no actual synonyms because I was hoping to implement a feature that would allow people to search by concept identifiers directly in NameRes. But until that functionality exists, we don't need to make the NameRes synonym files larger than they need to be.


gaurav:
> Not relevant to this PR exactly, but the fact that we need the same code in 2 places is kind of a smell.

Definitely! That has been filed as its own issue: https://github.com/TranslatorSRI/Babel/issues/126

Issue #158, opened 2023-06-26
reconsider name choices


cbizon:
At the moment we choose a preferred label by following the label of the preferred prefix.  Eg. if we're looking at a chemical, we take pubchem's label. But sometimes this leads to ugly names.   Perhaps we should find a way to choose a nicer name, e.g. https://github.com/NCATSTranslator/Feedback/issues/259#issuecomment-1605140850

gaurav:
Here is the priority order used by MolePro when choosing chemical names: https://github.com/broadinstitute/molecular-data-provider/blob/b13f566911ee8bf7a88361734c245ce9aa26f3b5/MoleProDB/builder/conf/sourcePriority.txt (thanks, @vdancik!)

gaurav:
See https://github.com/NCATSTranslator/Feedback/issues/568 for an example.

gaurav:
At least some of our long names are coming from PUBCHEM.COMPOUND. For example, PUBCHEM.COMPOUND:3420 has equivalent identifiers:

```json
      {
        "identifier": "PUBCHEM.COMPOUND:3420",
        "label": "4-Cyclohexyl-1-[2-[(2-methyl-1-propanoyloxypropoxy)-(4-phenylbutyl)phosphoryl]acetyl]pyrrolidine-2-carboxylic acid"
      },
      {
        "identifier": "CHEMBL.COMPOUND:CHEMBL4078476",
        "label": "CHEMBL4078476"
      },
      {
        "identifier": "CAS:1910773-95-3"
      },
      {
        "identifier": "HMDB:HMDB0252464",
        "label": "Fosenopril"
      },
      {
        "identifier": "INCHIKEY:BIDNLKIUORFRQP-UHFFFAOYSA-N"
      }
```

I think we should push PUBCHEM.COMPOUND below HMDB in the priority list.

PR #201, opened 2023-11-03
Babel v1.3 ongoing fixes

gaurav:
This PR collects changes we'd like to incorporate into Babel v1.3. As with previous releases, the goal is to review these commits as separate PRs, and hopefully then we can close this PR without merging it at all.

PR #210, opened 2023-12-04
Improve preferred names by boosting certain prefixes in choosing preferred name

### LINKED ISSUE #142: Hair preparations (UMLS:C0018503) does not have a preferred_label in Babel 2023may18
It really should.

gaurav:
This PR starts the process of improving preferred names, including:
- Synonym files used to use only the label of the most preferred identifier. This PR changes that so that it uses the label of the first identifier with a label. Should close #142.
- A first stab at implementing "boosted prefixes" configured by Biolink type, so that we can specify that for e.g. biolink:ChemicalEntity, we should boost certain prefixes when it comes to chooses preferred_name.
  - The idea for doing this comes from https://github.com/TranslatorSRI/Babel/issues/158#issuecomment-1769215304
  - Note that this will only set the preferred_name for NameRes, not NodeNorm, which calculates it on the fly. Changing that will either require adding the preferred label to the database or to duplicate this algorithm there. I prefer the former approach, but that's definitely something we'll need to consider in the next release.

The current preferred name boost prefixes are:

```
  "preferred_name_boost_prefixes": {
    "biolink:ChemicalEntity": [
      "DRUGBANK",
      "GTOPDB",
      "DrugCentral",
      "CHEMBL.COMPOUND",
      "RXCUI",
      "PUBCHEM.COMPOUND",
      "CHEBI",
      "HMDB"
    ]
  }
```


gaurav:
This ended up getting merged in PR #201. I'll move one minor update into PR #235 and close this PR.

PR #217, opened 2023-12-07
Fix conflation bug

gaurav:
This PR fixes the DrugChemical conflation bug (https://github.com/TranslatorSRI/NodeNormalization/issues/221) in a simple way that is probably the best we can do in the short term. The conflation code assumes that every identifier in the clique has a type, but that is not currently in the case for the drug-chemical conflation, which generates identifiers that aren't clique leaders. We can remove those by ensuring that every object must be in either `drug_rxcui_to_clique` or `chemical_rxcui_to_clique` -- if not, that object is skipped from the conflation. This gives us smaller conflations than we would otherwise get, but ensures that every identifier is a clique leader.

This has already been incorporated into PR #201, so if we decide not to go this route, we will need to remove it from that PR.


gaurav:
Superceded by https://github.com/TranslatorSRI/Babel/pull/266 which provides a more comprehensive solution -- closing.

Issue #116, opened 2023-03-17
Add PharmGKB mappings for genes, diseases, variants, and chemicals


gaurav:
As requested at https://github.com/TranslatorSRI/NodeNormalization/issues/171:

> To incorporate the PharmGKB API as described in https://github.com/biothings/BioThings_Explorer_TRAPI/issues/556, we need node normalizer to resolve PharmGKB identifiers for genes, diseases, variants, and chemicals. Examples of each:
>
> Gene: BRAF - https://www.pharmgkb.org/gene/PA25408
> Disease: leukemia - https://www.pharmgkb.org/disease/PA444750
> Variant: rs113488022 - https://www.pharmgkb.org/variant/PA166157522
> Chemical: imatinib - https://www.pharmgkb.org/chemical/PA10804
>
> Mappings can be found in the "Primary Data" section of https://www.pharmgkb.org/downloads
>
> Related issue to add prefixes to biolink model: https://github.com/biolink/biolink-model/issues/1236

gaurav:
Looking at [their downloads](https://www.pharmgkb.org/downloads), it looks like genes and chemicals (and drugs) have cross-mappings with identifiers already in NodeNorm, so those should be relatively easy to ingest. Variants and diseases are going to be trickier: variants don’t have cross-mappings with OMIM, which is the only source of variants currently in NodeNorm (I think), and diseases/phenotypes don’t have cross-mappings with MONDO and HP, which is what we most rely on (particularly for autocomplete). So I’m concerned about ingesting those without knowing how to merge it into existing variants and disease/phenotypes could get tricky.

gaurav:
A lot of the MESH mappings they have connect to cliques already in Babel (even if they don't always connect with MONDO), so adding these mappings shouldn't make things _worse_.

PR #219, opened 2023-12-11
Add PharmGKB

gaurav:
Adds identifiers from PharmGKB as per #116. WIP.


gaurav:
Closing — this draft has been idle for two years. PharmGKB still isn't in `src/datahandlers/` and `src/prefixes.py` only carries the prefix constant, so the underlying request (#116) is unresolved, but this branch is no longer a viable starting point. Reopening from a fresh branch on current main will be easier than rebasing 1100+ commits.

PR #248, opened 2024-03-18
Move DrugChemical synonyms into subdirectory

gaurav:
We currently create a DrugChemicalConflated.txt synonyms file using the DrugChemical conflation information to combine cliques. However, we still include synonym names for each individual chemical output (i.e. SmallMolecule.txt, Drug.txt, and so on). This PR moves those files into a subdirectory, so we still create those files, but they are in a separate directory so a user won't get confused and accidentally use both DrugChemicalConflated.txt and the redundant synonyms in the individual chemical output files.


gaurav:
This is harder to implement than I expected, because it would require modifying write_compendium() so it can write compendia and synonym files into different locations, which will be a lot of work. Closing branch.

PR #266, opened 2024-04-16
Fix DrugChemical conflation typing

### LINKED ISSUE #264: Water results in "WATER O 15" (PUBCHEM.COMPOUND:10129877) in NameRes because of a conflation issue
This is because NameRes entries are all based on DrugConflated results, and the conflation for water is:

```
["PUBCHEM.COMPOUND:10129877", "CHEBI:15377", "CHEBI:33813", "RXCUI:150985", "RXCUI:204918", "RXCUI:340584", "RXCUI:379002", "RXCUI:1043588", "RXCUI:1045437", "RXCUI:1045439", "RXCUI:1053147", "RXCUI:1053148", "RXCUI:1053172", "RXCUI:1053173", "RXCUI:1053428", "RXCUI:1053429", "RXCUI:1151100", "RXCUI:1151101", "RXCUI:1161792", "RXCUI:1161794", "RXCUI:1161795", "RXCUI:1180556", "RXCUI:1235498", "RXCUI:1235499", "RXCUI:1235500", "RXCUI:1235501", "RXCUI:1235502", "RXCUI:1235503", "RXCUI:1235504", "RXCUI:1310241", "RXCUI:1314884", "RXCUI:1423320", "RXCUI:1423321", "RXCUI:1424601", "RXCUI:1424602", "RXCUI:1424603", "RXCUI:1424604", "RXCUI:1424605", "RXCUI:1425974", "RXCUI:1425975", "RXCUI:1425976", "RXCUI:1425977", "RXCUI:1425978", "RXCUI:1489375", "RXCUI:1489376", "RXCUI:1489377", "RXCUI:1489378", "RXCUI:1539535", "RXCUI:1549855", "RXCUI:2108561", "RXCUI:2282752", "RXCUI:2282753", "RXCUI:2360606", "RXCUI:2360607", "RXCUI:2360608", "RXCUI:2360609", "RXCUI:2360610", "RXCUI:2601721", "RXCUI:2601722", "UMLS:C0359299", "UMLS:C3857954", "UMLS:C1883551"]
```

So why is PUBCHEM.COMPOUND:10129877 ("WATER O 15") ranked above CHEBI:15377 ("water")? This is because after we generate the initial conflation, the leading ID is RXCUI:1425974 ("Opticlear"), which is a biolink:Drug. As a biolink:Drug, PUBCHEM.COMPOUND is a more preferred prefix than CHEBI:

```
INFO:src.createcompendia.drugchemical:Leading ID RXCUI:1425974 normalized to RXCUI:1425974 (type biolink:Drug) with prefixes: ['ncats.drug', 'RXCUI', 'NDC', 'UMLS', 'PUBCHEM.COMPOUND', 'CHEMBL.COMPOUND', 'UNII', 'CHEBI', 'MESH', 'CAS', 'GTOPDB', 'HMDB', 'KEGG', 'KEGG.COMPOUND', 'ChemBank', 'PUBCHEM.SUBSTANCE', 'INCHI', 'INCHIKEY', 'KEGG.GLYCAN', 'KEGG.ENVIRON', 'SIDER.DRUG', 'BIGG.METABOLITE', 'foodb.compound']
```

So, options for fixing this:
1. Conflate fewer things together, so things like Opticlear won't get conflated with water. But this will be much trickier to implement.
2. We could try determining the type not by a random ID but by some sort of consensus calculation, but given all the RXCUIs I suspect everything will end up as a biolink:Drug.
3. I don't think we ever want RXCUIs to affect the type calculation. So we could filter them all out (along with all the biolink:ChemicalEntity CURIEs), then base the type on a consensus of the other IDs.
4. ???

gaurav:
We previously used a randomly chosen identifier from each DrugChemical conflation to choose the Biolink type for the entire conflation, which would also determine the order of prefixes within the conflation. This lead to issues where we used an RXCUI to determine that a conflation should be considered a biolink:Drug, when really biolink:SmallMolecule would be a better type. Instead, this PR replaces that approach with a preferred-type approach.

Closes #264.

WIP: currently being tested.


src/categories.py
```diff
@@ -9,7 +9,7 @@
 CELLULAR_COMPONENT = 'biolink:CellularComponent'
 CHEMICAL_ENTITY = 'biolink:ChemicalEntity'
 CHEMICAL_MIXTURE = 'biolink:ChemicalMixture'
-COMPLEX_CHEMICAL_MIXTURE = 'biolink:ComplexMolecularMixture'
+COMPLEX_MOLECULAR_MIXTURE = 'biolink:ComplexMolecularMixture'
```
cbizon, line 12: This is admittedly not exactly the envisioned use case but the idea of these consts was so that when a 1:1 change occurred here you could just change the string and save yourself many edits throughout the code.

cbizon (approved):

src/categories.py
gaurav: Yes, I know! But I took advantage of the centralized constants to bulk-rename one that had fallen out of sync with the corresponding Biolink type. If it had been a large change I would have proposed it in its own PR, but since it was only used in a handful of places and was related to the overall PR changes, I decided to incorporate it in here. For future changes, I'll update the string first and only update the constant name once it's pretty clear Biolink isn't going to change it again any time soon.

Issue #205, opened 2023-11-26
Come up with a plan for documenting Babel provenance/versioning


gaurav:
It may be possible to use Snakemake reporting to make this happen: https://snakemake.readthedocs.io/en/stable/snakefiles/reporting.html

Essentially, figure out how to embed either explicit version number (e.g. `Download UMLS release 2023AB`) or implicit version number (e.g. `Downloaded the latest version of XYZ as of Nov 26, 2023`) into the reports.

This may also help with Babel documentation (#148).

gaurav:
The only provenance information that Babel currently has is various "intermediate" files (e.g. https://stars.renci.org/var/babel_outputs/2024mar24/intermediate/), each of which includes the mappings that we use to construct our cliques. It is generally true that no mapping information outside the intermediate files is used, and it is generally true that every intermediate file comes from a single source. It is not guaranteed that a particular mapping is only present in a single file, and the intermediate files include identifiers that are left out of the final cliques because those prefixes are not allowed by the Biolink model.

This provides a potential way to provide both a quick-and-dirty and a more long term way of assembling metadata:
1. (Quick-and-dirty) We can manually come up with a JSON file that maps every intermediate file to its source.
2. (Long term) We can modify every Snakemake target that produces an intermediate file so that it produces a source-file that sits beside the intermediate file. Then, at the end, we can collect all the intermediate-metadata files into a single JSON file that maps each intermediate file to its source. Each Snakemake target can pull information from its corresponding download file to include metadata, such as version number, last modified time, and download time.
3. We load all of this mapping information into some kind of database.
4. (Quick-and-dirty) We can provide a provenance API; for a given ID, we can:
   1. Ask NodeNorm for the full clique.
   2. We query the database to find every mapping from the preferred ID to every secondary ID, either directly or via intermediary identifiers.
   3. We return the list of all mappings along with their provenance information.
5. (Long term) We incorporate the mapping information with provenance information into the cliques, store them in NodeNorm, and return them as needed.

gaurav:
The key to understanding Babel provenance at the moment are the concord files: these provide mappings from one (or more) identifier systems to other identifier systems, and it's generally true that every mapping in Babel originates in one or more of these files. So here is a simple data model I propose for slowly introducing provenance information into Babel:

[...]

PR #328, opened 2024-08-04
Adding SSSOM export

### LINKED ISSUE #33: SSSOM output
Instead of a special file format, we should perhaps be writing out SSSOM file.  

Is that going to make the files even more gigantic?  Probably?  Is that a problem?

gaurav:
This PR adds a SSSOM export to Babel. It is similar to [the code I wrote previously](https://github.com/TranslatorSRI/babel-validation/blob/84809c647a1e62f215606b6fe2c8b90c71886937/src/main/scala/org/renci/babel/utils/converter/Converter.scala#L209-L282), but is essentially a re-implementation. We might want to revisit this after https://github.com/TranslatorSRI/Babel/issues/205#issuecomment-2407522660 so we that we can include provenance information in the SSSOM output.

Closes #33, will help towards https://github.com/TranslatorSRI/NodeNormalization/issues/111.

WIP. Should be merged after #336.


gaurav:
Closing — this draft has been idle for 1.5 years. Its noted prerequisite #336 was merged in September 2024, but no SSSOM export has been added to main since (no `sssom` references in `src/`). The PR body also flagged that the provenance redesign might change the right approach. If SSSOM export is still wanted, easier to start fresh than rebase past 1500 commits.

PR #329, opened 2024-08-07
Cell wikidata

cbizon:
This hits wikidata to find CL/UMLS concordances via wikidata IDs.  It writes a concord table between CL and UMLS.

It could also bring in wikidata id's but not sure if that's terribly useful.  

Does a pretty good job; there are still lots of nodes that are either Cl or UMLS but cleans up many unmerged cliques.


gaurav (approved):
LGTM -- thanks, Chris!

PR #330, opened 2024-08-07
Reduce DrugChemical for load into SAPBERT

### LINKED ISSUE #313: Downscore labels which are very long or look like chemical names
We could skip those names entirely unless no other names are available, in which case we use the smallest chemical name.

gaurav:
This PR produces a DrugChemicalSmaller file so that we can load it into SAPBERT. It includes three changes:
1. I've tweaked the order of the prefix boosts so that we pick slightly better names.
2. A new `demote_labels_longer_than` config setting (currently set to 15) filters out any name longer than that size as long as at least one label equal to or less than that size is available.
3. Generate a SAPBERT training file called DrugChemicalConflatedSmaller.txt.gz which only includes cliques from DrugChemicalConflated if the preferred label is shorter than `DRUG_CHEMICAL_SMALLER_MAX_LABEL_LENGTH`.

`DRUG_CHEMICAL_SMALLER_MAX_LABEL_LENGTH` can be used to control how big DrugChemicalConflatedSmaller.txt is:
| DRUG_CHEMICAL_SMALLER_MAX_LABEL_LENGTH | Training rows | Unique CURIEs |
| ----------------------------------------------------- | ------------- | --------------- |
| 50 | 25,187,771 | 19,835,134 |
| 40 | 15,450,212 |  10,571,449 |
| 30 | 9,620,711 | 6,056,510 |
| 15 | 4,803,665 | 3,808,855 |

`DRUG_CHEMICAL_SMALLER_MAX_LABEL_LENGTH=30` seems like a reasonable setting right now.

Closes #313.


src/babel_utils.py
```diff
@@ -436,15 +436,17 @@ def write_compendium(synonym_list,ofname,node_type,labels={},extra_prefixes=[],i
...
                     #    the preferred CURIE order, but if this clique is in one of the Biolink classes in
                     #    preferred_name_boost_prefixes, we boost those prefixes in that order to the top of the list.
                     # 2. We filter out any suspicious labels.
                     #    (If this simple filter doesn't work, and if prefixes are inconsistent, we can build upon the
                     #    algorithm proposed by Jeff at
                     #    https://github.com/NCATSTranslator/Feedback/issues/259#issuecomment-1605140850)
-                    # 3. We choose the first label that isn't blank. If no labels remain, we generate a warning.
+                    # 3. We filter out any labels longer than config['demote_labels_longer_than'], but only if there is
+                    #    at least one label shorter than this limit.
+                    # 4. We choose the first label that isn't blank. If no labels remain, we generate a warning.
```
gaurav, line 449:
```suggestion
                    # 4. We choose the first label that isn't blank (that allows us to use our rule of smallest-prefix-first to find the broadest name for this concept). If no labels remain, we generate a warning.
```

PR #360, opened 2024-10-16
Added some additional manual concords

gaurav:
Some additional manual concords for Disease/Phenotype cliques and DrugChemical conflation.


cbizon (approved):
These look good.  I'm slightly uneasy about including the MONDO ones.  Can we make a PR or issue on MONDO to add these to their exact matches?

gaurav:
> These look good. I'm slightly uneasy about including the MONDO ones. Can we make a PR or issue on MONDO to add these to their exact matches?

Good point. I dug into this a bit deeper and discovered that these xrefs are actually present in MONDO _and_ in our intermediate files -- we're just separating the MONDO identifier into its own clique for some reason, which adding it to the manual concord in this PR does override. I'll merge this for now, but then I'll try to figure that out in https://github.com/TranslatorSRI/Babel/issues/369

PR #381, opened 2024-12-16
Add a comparator script

gaurav:
This PR adds a comparator script that can be used to compare compendium files between different Babel runs. The idea is to support a development process where:
1. You produce a Babel output (output1).
2. You modify the code in some way.
3. You produce a modified output (output2).
4. You run this script to produce a report that lets you know how your code has changed the outputs.

WIP. Should be merged after PR #365.


cbizon (approved):
Looks good

PR #395, opened 2025-01-09
Goumls

cbizon:


cbizon:
This adds UMLS/GO mappings from UMLS.  This happens in both the Process/Function and Anatomy (CellularComponent) codes.

Some points of note:
1. UMLS is not currently a valid type for BiologicalProcess, so there is a hack in node.py to handle this case until my PR gets merged
2. The main annoyance here is that the UMLS/GO mappings include lots of obsolete GO terms.    We filter out the ones that UMLS knows are obsolete, but then we also check at glom-time against the list of GO ids and reject any mappings that are not in our GO list.
3. This will move some identifiers / STYs out of the UMLS leftover stuff.   I am now realizing that I don't know if that's built by looking at the UMLS id's we didn't map or by pulling particular STYs.  If the latter then I guess that there's a bit more work to do so that we don't double pull.

cbizon:
OK, I checked on (3) and it seems like this won't be a problem

gaurav (approved):
LGTM, thanks for writing this!

PR #396, opened 2025-01-13
Efo

cbizon:
Trying to create a bit more cohesion specifically in phenotypic features.

Reviewing Bagel results, we found that EFO was often unlinked, especially for measurements (e.g. Age).  

Made 2 basic changes
1. EFO has a lot of skos:ExactMatch edges, but not for every EFO.  There are lots of entries where there are only dbXrefs.  Now, we pull in dbXref matches for EFO terms, but ONLY for terms that don't have exactMatch terms.  The thought being that if we have some as marked with exact then things that aren't explicitily exact probably are not.  But if we don't have anything that is exact, then dbXrefs are our best guess
2. Brought in one more UMLS class.  (OrganismAttribute).  This brings in things like "Age" which are coming in from EFO anyway.


gaurav (approved):
LGTM!

PR #419, opened 2025-03-05
Clo

cbizon:
Adds CellLine as a type and CellLineOntology (CLO) to Babel.  CLO doesn't contain equivalences to anything so this is really just getting all the CLO descendents of CLO:0000001 and pushing them through the machinery to generate the files.


gaurav (approved):
Looks good, and seems to work correctly!

PR #420, opened 2025-03-06
simple fma addition

cbizon:
This PR adds FMA to anatomical entity.  It got added to biolink this morning, but I guess we need to update something in babel to make this get picked up (wait for a new bmt?) 

@gaurav how does updating biolink in babel work?


cbizon:
Converted to draft until we can test it

gaurav:
> @gaurav how does updating biolink in babel work?

You should only need to modify the version in config.yaml: https://github.com/TranslatorSRI/Babel/blob/3d656dad89f07d54f3e244214e865f51ee3debc5/config.json#L7

If there's a change in the Biolink Model Toolkit, you may need to use pip to upgrade that.

cbizon:
I guess my changes haven't made it into a release yet, but are just merged into main.  So I guess I need to just wait for that?

PR #445, opened 2025-05-19
Moved UMLSToBiolinkTypeConverter into umls.py

gaurav:
This is to ensure that these type mappings can be used elsewhere in the source code if needed.

WIP: should we refined after #750


gaurav:
Closing — WIP draft pending issue #750, which is still open. The `UMLSToBiolinkTypeConverter` consolidation can be redone on top of whatever shape #750's resolution settles on; keeping a 1000+ commit-behind branch open as a placeholder isn't useful.

PR #459, opened 2025-07-07
Add metadata

gaurav:
This PR adds metadata.yaml files throughout the process (as suggested in https://github.com/TranslatorSRI/Babel/issues/205#issuecomment-2270334885):
- Final metadata files are stored in `babel_outputs/metadata`.
- These are based on the concord files, which will be in `babel_outputs/intermediate/*/concords/metadata-FILENAME.yaml`.
- These are based on download files, which will be in `babel_downloads/*/metadata.yaml`, but these have mostly not been implemented yet.

The actual metadata here is very ill-defined because I'm making it up as I'm going along, but eventually this will have a little bit of structure and later we can really lock this into a fixed, validatable format.


src/createcompendia/anatomy.py
```diff
@@ -91,14 +92,43 @@ def write_umls_ids(mrsty, outfile):
...
+                           sources=[
+                               {
+                                   'type': 'UberGraph',
+                                   'name': 'GO'
+                               }
+                           ],
+                           description=f'get_subclasses_and_xrefs() of {GO}:0005575'
+                           )
+    # TODO: delete
+    write_concord_metadata(metadata_yamls['CL'],
```
gaurav, line 125: - [ ] We need to add some more details here.

src/datahandlers/pantherfamily.py
```diff
@@ -38,3 +39,15 @@ def pull_labels(infile,outfile):
                 #labels[sub_family]=sfname
                 labelf.write(f'{sub_family}\t{sfname}\n')
                 done.add(sf)
+
+    write_metadata(
+        metadata_yaml,
+        typ='transform',
+        name='HGNC Gene Family labels',
```
gaurav, line 46: - [ ] Is this really HGNC Gene Family labels?

PR #473, opened 2025-08-03
Add support for properties and use them to store ChEBI secondary IDs

### LINKED ISSUE #86: Missing CHEBIs
There seem to be a lot of missing CHEBIs as found through Data Services...
Some examples:

```yaml
babel_tests:
  Resolves:
    - CHEBI:12457
    - CHEBI:86461
    - CHEBI:23770
    - CHEBI:12146
    - CHEBI:1037
    - CHEBI:23649
    - CHEBI:12894
    - CHEBI:572
    - CHEBI:1614
    - CHEBI:9253
    - CHEBI:26504
    - CHEBI:24774
    - CHEBI:76940
    - CHEBI:22051
    - CHEBI:22421
    - CHEBI:714
    - CHEBI:10968
    - CHEBI:3298
    - CHEBI:20232
    - CHEBI:27172
    - CHEBI:74908
    - CHEBI:19659
    - CHEBI:1216
```

There seem to be about 20000 of them.  Spot checking a handful, these seem to be cases where maybe the CHEBI has a structure but doesn't appear in unichem.   This doesn't seem to be a new problem, as it's true in both dev and prod.

----
As of Feb 21, 2026, the following CURIEs don't resolve:

```yaml
babel_tests:
  Resolves:
    - CHEBI:76940
    - CHEBI:22421
    - CHEBI:1216
```

gaurav:
This PR adds support for ChEBI secondary IDs to Babel cliques. This PR adds a new `properties.py` Python package that:
1. Defines a `Property` data class with a CURIE, predicate, value and sources (as a list). The predicates must be from a limited list of supported predicates (currently only `oboInOwl:hasAlternativeId`).
2. Methods for writing properties into gzipped JSONL files, and
3. A `PropertyList` that can be used to load and query information from multiple property in memory. Since properties are still pretty small, I don't worry about storing them all in memory.

It then uses this property system to store ChEBI secondary IDs (using the `oboInOwl:hasAlternativeId` property), and modifies write_compendia() so that any CURIE that has `oboInOwl:hasAlternativeId` property values will have those values (assumed to be CURIEs) added to the clique **after** the original CURIE in the clique, without affecting the type calculation. This allows us to add alternative IDs while making sure they will never be the preferred CURIE for a clique nor will affect its type, while still allowing it to pull in new synonyms or to be used in node normalization.

This code ended up being broken in ways I couldn't tell until I was able to fully finish a Babel run (i.e. after PR #482). Those are fixed in PR #483 by turning `Property` into a frozen dataclass, replacing the single source with a single source as a string, adding some example code for testing, and makes some incidental fixes to the metadata YAML files.

Note that this means that care must be taken to ensure that HAS_ALTERNATIVE_IDs are NOT included in any other clique, because other they will be duplicated! This is safe to do with ChEBI secondary IDs, because they are not included in the main ChEBI IDs.


cbizon:
The PR looks good to me.  Should things like IC use this property system?

PR #503, opened 2025-08-28
Add CLI normalizer

gaurav:
We have several requests/needs for files containing CURIEs to be normalized in bulk. There are multiple ways of doing this using NodeNorm, but it would be nice to have something that can do it as an INNER JOIN against the combined DuckDB database we create while building NodeNorm, as that should be way faster than other approaches. We could also use this to export every mapping we have from a particular source (i.e. https://github.com/TranslatorSRI/NodeNormalization/issues/321).

WIP, should be merged after #683 and #671 start a new generation of scripts within Babel.


gaurav:
Stale-branch sweep note: this draft is blocked on #683 and #671, both of which are themselves still open WIPs and haven't moved in months. The DuckDB schema and reporting infrastructure in main has shifted since this branch was written (834 commits behind). When the upstream report-generation work lands, it may be cleaner to redesign on top of that than to rebase this branch forward. Leaving open for now.

PR #508, opened 2025-08-30
Conditionally apply overused xref filtering in combine_unichem

cbizon:
## Summary

• Modified `combine_unichem` function to only call `remove_overused_xrefs` for specific prefixes
• Now only applies filtering to UNII, KEGG.COMPOUND, and DrugCentral prefixes
• Preserves all cross-references for other prefixes like CHEBI without filtering

## Background

We analyzed all the UNICHEM mappings, and the code and results can be found here: https://github.com/cbizon/UniChemCheckup/

Based on this analysis, we determined that overused cross-reference filtering should only be applied to certain prefixes that showed problematic patterns, while maintaining complete mappings for others.

## Test plan

- [x] Verify the function correctly detects prefixes from input files
- [x] Confirm filtering is only applied to UNII, KEGG.COMPOUND, and DrugCentral
- [x] Ensure other prefixes bypass the filtering step

### DIFF
--- a/src/createcompendia/chemicals.py
+++ b/src/createcompendia/chemicals.py
@@ -16,6 +16,8 @@ from src.babel_utils import write_compendium, glom, get_prefixes, read_identifie
 
 import src.datahandlers.mesh as mesh
 import src.datahandlers.umls as umls
+from src.util import Text
+
 
 def get_type_from_smiles(smiles):
     if '.' in smiles:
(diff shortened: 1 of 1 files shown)


gaurav (approved):
LGTM! Although it's interesting to me that Claude assumed that our prefixes had ':' suffixes, which we don't! I made a few changes and made the code a bit more paranoid in case we ever have a UNICHEM file with multiple prefixes.

PR #564, opened 2025-09-09
Bulk normalize

gaurav:
This PR adds a new feature called "bulk normalize", which downloads files stored on a particular directory on stars.renci.org, normalizes them against the currently built Babel, and produces a report on how many of those identifiers have been normalized. This is intended to be used for the use-case where we want to ensure that a large number of identifiers from a particular source can be normalized, and tracks our progress towards that goal.

This PR also fixes the case where RXCUI is included in ChemicalEntity preferred prefix list -- we delete it from wherever it is before adding it to the end.


gaurav:
Stale-branch sweep note: this PR's body mentions it also fixes the case where `RXCUI` is in the `ChemicalEntity` preferred-prefix list. As of current main, I don't see that fix elsewhere — `git log main --grep=RXCUI` doesn't show a deduplication change since this PR opened, so that part is still load-bearing. Suggest either rebasing forward or extracting the RXCUI fix into a separate PR before letting the branch age out further. Leaving open for now.

PR #565, opened 2025-09-09
Eliminate duplicate UMLS type warnings

gaurav:
This PR fixes duplicate log/report entries in leftover_umls.py. Because the same CURIE can appear multiple times in MRCONSO, warnings and report lines for NO_UMLS_TYPE and MULTIPLE_UMLS_TYPES were being emitted once per MRCONSO row rather than once per CURIE. The fix tracks seen CURIEs in sets  and skips repeated log/report writes. It also migrates all logging.xxx calls to a module-level `logger = get_logger(__name__)`.

Also includes some unrelated `uv.lock` version changes and some Snakemake file reformatting that really should have been fixed elsewhere >.<

### DIFF
--- a/src/createcompendia/leftover_umls.py
+++ b/src/createcompendia/leftover_umls.py
@@ -119,6 +119,10 @@ def write_leftover_umls(metadata_yaml, compendia, umls_labels_filename, mrconso,
                 if umls_id in umls_ids_in_this_compendium:
                     logger.debug(f"UMLS ID {umls_id} has already been included in this compendium, skipping.")
                     continue
+                if umls_id in curies_no_umls_type or umls_id in curies_multiple_umls_type:
+                    # This CURIE was already evaluated and skipped due to type resolution failure.
+                    # Skip it here to avoid redundant type lookups on subsequent MRCONSO rows for the same CUI.
+                    continue
 
                 # The STR value should be the label.
                 label = x[14]


src/createcompendia/leftover_umls.py
```diff
@@ -275,4 +279,4 @@ def umls_type_to_biolink_type(umls_tui):
 
         logging.info(f"Wrote out metadata file {metadata_yaml}.")
 
-    logging.info("Complete")
+    logging.info("Complete")
```
gaurav:
```suggestion
    logger.info("Complete")
```

src/createcompendia/leftover_umls.py
```diff
@@ -158,14 +157,18 @@ def umls_type_to_biolink_type(umls_tui):
                     biolink_types = [FOOD]
 
                 if len(biolink_types) == 0:
-                    logging.debug(f"No UMLS type found for {umls_id}: {umls_type_results} -> {biolink_types}, skipping")
-                    reportf.write(f"NO_UMLS_TYPE [{umls_id}]: {umls_type_results} -> {biolink_types}\n")
-                    count_no_umls_type += 1
+                    # We skip this CURIE, but we don't want to print multiple warnings for the same CURIE.
```
gaurav:
```suggestion
                    # We skip this CURIE, but we don't want to print multiple log messages for the same CURIE.
```

PR #588, opened 2025-09-25
Feat/use rust for datacollect

gaurav:


gaurav:
Stale-branch sweep note: this draft has no description, has been idle since September, and is 700+ commits behind main. A Rust rewrite of `datacollect` is a large architectural pivot that probably warrants a fresh design discussion if/when revisited — at which point starting over may be easier than rebasing. Flagging for eventual cleanup; leaving open for now.

PR #589, opened 2025-09-25
Improve Biolink type reporting

gaurav:
WIP

- [ ] Fix case where the list of Biolink types is incomplete
  - We use `STRING_AGG(cliques.biolink_type, '||' ORDER BY cliques.biolink_type ASC) AS biolink_types` to get a list of all the Biolink types (and also double-check the CURIE count). This appears to work for every row except for UniProtKB.
  - [ ] It would be nice to highlight the cases where it doesn't work (i.e. where the number of Biolink types is not equal to the CURIE count).
  - [ ] There must be some way of counting this as a subquery.


gaurav:
Closing — draft from September with an explicit TODO checklist about an unsolved subquery problem; the work hasn't progressed and the underlying DuckDB reporting code has shifted significantly in main since then. Easier to re-attempt against the current schema than rebase.

PR #670, opened 2026-02-23
Exclude UMLS/MeSH chemical identifiers when preparing proteins

gaurav:
This PR excludes UMLS, MeSH and DrugBank chemical identifiers when preparing proteins if those identifiers are already present in chemicals. But #692 may provide a more thorough solution. So I'll mark it as a draft for now.

### DIFF
--- a/src/createcompendia/protein.py
+++ b/src/createcompendia/protein.py
@@ -6,7 +6,7 @@ import src.datahandlers.umls as umls
 from src.babel_utils import Text, glom, read_identifier_file, write_compendium
 from src.categories import PROTEIN
 from src.metadata.provenance import write_concord_metadata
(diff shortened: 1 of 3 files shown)


gaurav:
Closing — the PR description already noted that #692 might be a more thorough solution. PR #692 was merged on 2026-04-27, along with #687 ("Eliminate MeSH proteins from chemicals") and the related issue #675 work, all of which now sit in `main`. The protein/chemical identifier separation is handled there in a more comprehensive way than this branch's targeted exclusion in `umls.build_sets()`. `fix-mesh-filtering` is 339 commits behind main and rebasing it onto the current code would essentially mean rewriting it on top of work that has already solved the problem. Deleting the branch.

PR #703, opened 2026-03-30
Reduce peak RSS of cross-compendium DuckDB checks via two-pass queries

gaurav:
The three cross-compendium report rules (check_for_identically_labeled_ cliques, check_for_duplicate_curies, check_for_duplicate_clique_leaders) previously requested 1500G of SLURM RAM. The bottleneck was LIST() aggregations over all Edge/Clique Parquet files — DuckDB cannot spill LIST() state to disk, so every growing array stayed in RAM.

Replace each single-pass LIST() query with a two-pass approach:
- Pass 1: pure COUNT(*) GROUP BY to find duplicate keys; fully spill-able
- Pass 2: LIST() collected only for the confirmed duplicates (a small join against the pass-1 result), keeping list materialization trivial

Also restores the biolink_type and clique_identifier_count columns in check_for_duplicate_clique_leaders that were previously dropped with a comment saying they used too much memory.

Lower Snakemake mem= and DuckDB memory_limit conservatively; right-size further after observing max_rss from benchmark TSVs on a real run.


gaurav:
Closing this without merging — its substance has already landed on `main` (via #860 and follow-ups) and gone further.

I merged the current `main` into this branch (`fix-large-duckdb-queries`, now at `1c4b729e`) and the result has a **zero net diff** against `main`. The merge conflicted in exactly the three files this PR touches, which is the tell: `main` already rewrote all of them.

Function by function:

- **`check_for_duplicate_curies`** and **`check_for_duplicate_clique_leaders`** — `main` already uses this PR's two-pass pattern (spillable `COUNT` `GROUP BY` first, `LIST()` only over confirmed duplicates), essentially the same SQL, plus `log_duckdb_settings_on_error` wrappers and a tuned `max_temp_directory_size`. `clique_leaders` also already restores `biolink_type` / `clique_identifier_count`.
- **`check_for_identically_labeled_cliques`** — `main` deliberately moved *past* this PR's approach. This PR groups on `LOWER(preferred_name)` and builds `LIST(... ORDER BY ...)`, which OOM-killed on the full graph (one name shared by ~92k cliques; `LIST()` isn't spillable, and the global sort overshot the cgroup limit). `main` now groups on `hash(LOWER(preferred_name))` and streams unsorted pairs via a hash join. This PR's version would be a regression.
- **Snakefile / README resources** — this PR's `mem` reductions (128–256G, `threads=4`) predate two failures since diagnosed on `main`: an `mmap`-count limit (`vm.max_map_count`, #846) that forces `identically_labeled_cliques` down to `memory_limit` 16G / 1 thread, and a `threads > 1` background-thread DuckDB OOM. Re-adopting `threads=4` / lower mem here would undo those fixes.

The one genuinely useful idea not yet on `main` — pinning `cpus_per_task` to each report rule's DuckDB `threads` — only matters for `check_for_duplicate_clique_leaders` (the one rule that runs `threads=4`), and it becomes load-bearing precisely when #869 drops the default `cpus_per_task` to 1. I'm folding that one-line change into #869 rather than carrying this branch.

The merged branch is pushed for the record; the original pre-merge state is preserved locally on `backup/fix-large-duckdb-queries-pre-merge`.

PR #778, opened 2026-05-18
Add src/predicates.py to centralize RDF predicate URI constants

gaurav:
Replaces hardcoded predicate strings (oboInOwl#hasExactSynonym, etc.) scattered across 7 data handler files and node.py with named constants imported from a new src/predicates.py module, mirroring the pattern already established by src/prefixes.py for CURIE prefixes.


src/prefixes.py
```diff
@@ -1,5 +1,6 @@
+# Prefixes used for CURIEs
+#
 # The point here is to keep prefixes in one place so that we can easily deal with things like PUBCHEM->PUBCHEM.COMPOUND
```
gaurav, line 3:
```suggestion
# The point here is to keep prefixes in one place so that we can easily deal with things like PUBCHEM->PUBCHEM.COMPOUND
# In principle, we should generate this off of the context file?
```

Issue #733, opened 2026-04-23
New data sources, EMAPA & MP


kevinschaper:
### What are you requesting?

- [x] Identifier system
- [ ] Mapping source

### Priority

High

### Data source name

Mouse Phenotype Ontology, Mouse Developmental Anatomy Ontology

### Data source URL

https://informatics.jax.org/vocab/mp_ontology/ https://obofoundry.org/ontology/emapa.html 

### Biolink type of these identifiers

biolink:PhenotypicFeature, biolink:AnatomicalEntity

### Value of including this source

For translator's alliance ingest (gene to phenotype, gene expression), we're getting normalization misses on 3889 EMAPA terms and 11,105 MP terms. I think it's likely that these terms will all be clique leaders or unique for the concepts they represent, so hopefully that makes it an easier one to load.  (I'm assuming that you won't normalize to uberon or upheno, and will just keep the species specific terms.)

### Potential conflicts with existing sources

Shouldn't conflict with any existing sources

gaurav:
> (I'm assuming that you won't normalize to uberon or upheno, and will just keep the species specific terms.)

@kevinschaper Wouldn't it make more sense to combine these cliques, using e.g. the cross-references at https://monarchinitiative.org/UBERON:0012313 to combine [UBERON:0012313](http://purl.obolibrary.org/obo/UBERON_0012313) with [EMAPA:16389](http://purl.obolibrary.org/obo/EMAPA_16389)? One reason for doing that is because otherwise you'll end up with both terms if you try searching NameRes for "1st arch maxillary ectoderm". However, when I last tried adding MP (https://github.com/NCATSTranslator/Babel/pull/300), I wasn't able to find comprehensive cross-references: I ended up with 13,335 MP terms that didn't combine with anything else. But if you think that's okay, that would certainly simplify adding these identifiers!

We could also add "(mouse)" to all the EMAPA/MP terms to disambiguate them.

cbizon:
[...]

Issue #730, opened 2026-04-22
Pipeline tests reveal MeSH duplication between different compendia


gaurav:
Two different representations:

```
        # protein / anatomy overlaps
        "MESH:D022041", "MESH:D035321", "MESH:D006570", "MESH:D009707",
        "MESH:D064448", "MESH:D035341", "MESH:D045524", "MESH:D007106",
        "MESH:D000067816", "MESH:D000089804", "MESH:D002843", "MESH:D017358",
        "MESH:D008894", "MESH:D000961", "MESH:D009360",
        # chemicals / anatomy overlaps
        "MESH:D000091083", "MESH:D014688",
        # anatomy / diseasephenotype overlaps
        "MESH:D000072717", "MESH:D018404", "MESH:D065309", "MESH:D003809",
        "MESH:D008467", "MESH:D012303", "MESH:D017439", "MESH:D000153",
        "MESH:D014097", "MESH:D010677", "MESH:D000072662", "MESH:D003750",
        "MESH:D008551", "MESH:D048629", "MESH:D002921", "MESH:D007627",
        # anatomy / taxon overlaps
        "MESH:D036226", "MESH:D013172", "MESH:D052940", "MESH:D033761",
        "MESH:D053058", "MESH:D002523", "MESH:D038821", "MESH:D033661",
        "MESH:D013171", "MESH:D034101", "MESH:D013104", "MESH:D052939",
        "MESH:D013170",
```

And:

```
[...]

Issue #729, opened 2026-04-22
Pipeline tests reveal UMLS duplication between different compendia


gaurav:
These are:
* UMLS:C5443441: ['chemicals', 'anatomy']
* 'UMLS:C5443442': ['chemicals', 'anatomy']

PR #781, opened 2026-05-20
Add EMAPA as an anatomy source

gaurav:
## Summary

Adds EMAPA (Mouse Developmental Anatomy Ontology) as an anatomy identifier source (#733), contributing **8,078 identifiers** to the anatomy compendia. Source documentation, the generated impact report and a full build-vs-build clique diff are under [`docs/sources/EMAPA/`](docs/sources/EMAPA/).

Two things for an SME to weigh in on are at the bottom; everything above them is evidence for those two decisions.

## What this does to existing cliques: nothing

[`docs/sources/EMAPA/clique-diff.md`](docs/sources/EMAPA/clique-diff.md) records a full `babel-clique-diff` between anatomy built without and with EMAPA, from identical cached intermediates on the same commit. It closes the gap the impact report cannot see (#895): before-cliques that split, lose members, or disappear.

**No pre-existing clique changes in any way.** Zero change rows: nothing splits, nothing is retyped, nothing loses a member, and no identifier is dropped from any of the four anatomy compendia. Every difference is a brand-new EMAPA-only clique.

| compendium | before | after | diff |
|---|---|---|---|
| `AnatomicalEntity.txt` | 145,743 | 147,565 | +1,822 |
| `Cell.txt` | 9,197 | 9,197 | 0 |
| `CellularComponent.txt` | 9,469 | 9,469 | 0 |
| `GrossAnatomicalStructure.txt` | 10,708 | 12,639 | +1,931 |

Totals go 175,117 → 178,870; the +3,753 is exactly the pure-new clique count the impact report computes independently.

## How EMAPA is represented in Babel

Every EMAPA term is typed at extraction: terms at or below [`EMAPA:35949`](http://purl.obolibrary.org/obo/EMAPA_35949) "organ" or [`EMAPA:35868`](http://purl.obolibrary.org/obo/EMAPA_35868) "tissue" as `biolink:GrossAnatomicalStructure`, everything else as `biolink:AnatomicalEntity`.

That declared type is only a starting point — the **clique's** type wins, and EMAPA is placed last in the type-precedence order, behind GO, CL and UBERON. So EMAPA can add typing to a clique that had none, but can never override an existing ontology's judgement:

| | declared by EMAPA | final, in the compendia |
|---|---|---|
| `biolink:AnatomicalEntity` | 4,090 | 2,787 |
| `biolink:GrossAnatomicalStructure` | 3,988 | 5,291 |

The totals match at 8,078, so the ~1,300-term shift is retyping — an EMAPA term called an anatomical entity landing in `GrossAnatomicalStructure.txt` because a UBERON member of its clique is typed that way — not loss. No EMAPA identifier is dropped.

**How EMAPA joins to the rest of anatomy.** EMAPA asserts no outgoing mapping triples at all in UberGraph — no `hasDbXref`, no `skos:exactMatch`/`closeMatch`/`broadMatch`/`narrowMatch`. Every EMAPA↔UBERON equivalence comes from the other direction: UBERON asserts 4,356 `hasDbXref` triples pointing at EMAPA. So the identifiers EMAPA contributes are the substance of this ingest, and the mappings that place them into cliques are UBERON's.

## Fixes two cliques that merged a gross structure with a cell type

Found by scanning for identifiers in a compendium they had no business being in — an unexpected Biolink type is the cheapest signal that a merge went wrong. **Both conflations predate this PR and ship on `main` today.**

- [`CL:0000166`](http://purl.obolibrary.org/obo/CL_0000166) "chromaffin cell" shared a `biolink:Cell` clique with [`UBERON:0001236`](http://purl.obolibrary.org/obo/UBERON_0001236) "adrenal medulla" — the cell type merged with the structure it sits in. Cause: `UBERON:0001236 hasDbXref MESH:D019439`, and MESH:D019439 is "Chromaffin Cells".
- [`UBERON:0003124`](http://purl.obolibrary.org/obo/UBERON_0003124) "chorion membrane" sat in a `biolink:CellularComponent` clique led by [`GO:0042600`](http://purl.obolibrary.org/obo/GO_0042600) "egg chorion" — a mammalian extraembryonic membrane merged with the acellular envelope of an insect egg. Cause: `UMLS:C0008503 eq GO:0042600` in the UMLS concord.

Neither is fixable at the prefix level (MESH and GO are legitimate anatomy xref targets), so both pairs are now listed individually in `input_data/anatomy_badxrefs.txt`.

**This also rescues two EMAPA terms the branch was silently losing.** [`EMAPA:18428`](http://purl.obolibrary.org/obo/EMAPA_18428) "adrenal medulla" and [`EMAPA:16112`](http://purl.obolibrary.org/obo/EMAPA_16112) "chorion" were dragged into those mistyped cliques and then dropped entirely, because EMAPA is not a permitted prefix for `biolink:Cell` or `biolink:CellularComponent`. A clique diff cannot see this class of loss — a dropped identifier appears on neither side.

### 1. EMAPA identifiers are not forced to be unique within a clique

122 UBERON terms cross-reference more than one EMAPA term (263 EMAPA CURIEs); for 116 of them, more than one target is a live, labelled EMAPA term. Full list, one row per mapping, in [`docs/sources/EMAPA/multi-emapa-xrefs.csv`](docs/sources/EMAPA/multi-emapa-xrefs.csv).

Reading that list, these are overwhelmingly **genuine 1:n mappings** — a mouse developmental partonomy naming a structure at finer or stage-specific granularity than the adult UBERON term:

| UBERON term | EMAPA terms it xrefs |
|---|---|
| `UBERON:0000044` "dorsal root ganglion" | `EMAPA:16667` "spinal ganglion", `EMAPA:16668` "dorsal root ganglion", `EMAPA:18372` "tail dorsal root ganglion" |
| `UBERON:0001017` "central nervous system" | `EMAPA:16470` "central nervous system", `EMAPA:16754` "tail central nervous system" |
| `UBERON:0001049` "neural tube" | `EMAPA:16164` "neural tube", `EMAPA:16757` "tail neural tube" |

Restricting EMAPA to one identifier per clique would block all 122 merges, and would have silently withdrawn three already-published identifiers from the compendia.

- [x] **Confirm** that collapsing e.g. "spinal ganglion", "dorsal root ganglion" and "tail dorsal root ganglion" into one clique with `UBERON:0000044` is acceptable. Restricting EMAPA does not produce a *better* clique — it produces an arbitrary winner plus singletons, and can silently delete identifiers. If these mappings should not be equivalences at all, the fix is to filter the offending UBERON xrefs, not to restrict the prefix.

### 2. EMAPA terms carry no taxon

We considered attaching [`NCBITaxon:10090`](http://purl.obolibrary.org/obo/NCBITaxon_10090) "Mus musculus" to every EMAPA term, since EMAPA is by definition a mouse ontology, and **decided against it**. The tempting precedent is MP, where the taxon cleanly separates MP (mammal) from HP (human). EMAPA is not analogous: of the 8,078 EMAPA CURIEs in the built compendia, 4,345 share a clique with a UBERON term, and UBERON is a multi-species ontology. Marking those cliques mouse-specific would be incorrect — `UBERON:0001049` "neural tube" is not a mouse structure just because `EMAPA:16164` is one of its cross-references. A taxon attaches to the clique, not to the EMAPA identifier within it, so there is no way to say "mouse, but only for the EMAPA member".

- [x] **Confirm** that no taxon is preferable to a mouse taxon here.

Both decisions are recorded in [`docs/sources/EMAPA/README.md`](docs/sources/EMAPA/README.md#two-decisions-that-need-sme-confirmation).

## Implementation notes

For code review; none of this changes the answers above.

- EMAPA is a `part_of` partonomy rather than an is_a hierarchy, so `build_sets()` and `get_subclasses_of()` now take a `hierarchy_predicate` (defaulting to `rdfs:subClassOf`). A `subClassOf` walk from `EMAPA:0` reaches only two terms.
- `build_sets()` now sorts its output. It previously iterated a set of strings, so identical code and data produced different concord row orders — and therefore different cliques — between runs (#945). EMAPA ended up unrestricted so this decides no EMAPA clique, but UBERON and GO are restricted and the guarantee makes any such tie-break reproducible.
- `input_data/anatomy_badxrefs.txt` is applied through the `concord_pair_filter` hook `glom_from_files()` already exposes, so it needed no new plumbing.

## Test plan

- [x] `ruff check` / `ruff format` / `snakefmt` / `rumdl` — all clean
- [x] `uv run pytest -m unit -q` — 421 passed
- [x] `uv run pytest tests/pipeline -q --all` — 56 passed, 3 skipped, 2 xfailed. Both xfails are pre-existing `KNOWN_DUPLICATES` entries (MESH → #730, UMLS → #729); EMAPA passes.
- [x] Impact report regenerated from a local `anatomy` build, after deleting the previous report entirely so every file provably comes from the new build
- [x] Clique diff regenerated from a deterministic two-sided build — 0 change rows
- [x] SME review of the two decisions above


docs/sources/EMAPA/README.md
```diff
@@ -0,0 +1,27 @@
+# EMAPA source notes
+
+EMAPA (Mouse Developmental Anatomy Ontology) is used in Babel as an anatomy source.
```
gaurav:
```suggestion
EMAPA (Mouse Developmental Anatomy Ontology, https://obofoundry.org/ontology/emapa.html) is used in Babel as an anatomy source.
```

gaurav:
Heads-up after rebasing onto the updated `add-emapa` (#742): the shared clique-building skeleton was refactored out of anatomy into a new module `src/createcompendia/cliques.py` (commit `154f3788`).

`anatomy.compute_cliques_for_impact_report` is now a thin wrapper over `cliques.glom_from_files(...)`, which runs the common `load ids → glom`, `load concords → filter → drop overused xrefs → glom` loop parameterized by three hooks:

- `concord_pair_filter(parts, infile, dicts)` — per-pair keep/drop, with access to the clique state built so far
- `overused_xref_remover(pairs, infile)` — per-file `remove_overused_xrefs` variant
- `glom_kwargs` — extra kwargs forwarded to every `glom` call (e.g. disease's `close={MONDO: ...}`)

`build_compendia` still calls the wrapper, so the impact report's reglom stays in lockstep with the real build (no drift).

This `AddingNewSources.md` was written before that refactor, so two sections describe the compute hook as anatomy-only logic and will be slightly stale once rebased:

- **§3 "Add a compendium-building hook"** — should describe a new type's `compute_cliques_for_impact_report` as a thin wrapper supplying its hooks to `cliques.glom_from_files`, rather than reimplementing the glom loop.
- **§Future "Roll out the compute hook to other semantic types"** — can note that the skeleton already exists in `cliques.py`, so rolling out to disease/taxon/process is now adding a thin wrapper + a `SEMANTIC_TYPE_CONFIG` entry (and, ideally, routing each type's real `build_compendia` through the same wrapper, or adding a sync test).

The CLAUDE.md update in #742 (commit `25174094`, "extending the report to a new semantic type" section) has ready-to-lift wording for both.

cbizon:
Can this be merged?

docs/sources/EMAPA/README.md
```diff
@@ -0,0 +1,32 @@
...
+  `part_of` traversal works.
+- `mappings.md` — how EMAPA cross-references to other prefixes are extracted from UberGraph
+  xrefs and written as concord rows.
+- `impact-report.md` — auto-generated report quantifying the identifiers, biolink types, and
+  clique changes EMAPA adds to the anatomy compendium. See "Adding a new data source" in
+  `CLAUDE.md` for how to regenerate it.
+- `clique-diff.md` — a full build-vs-build diff of the anatomy compendia with and without EMAPA,
+  covering the restructured and dropped cliques the impact report cannot see. Explains why
+  adding EMAPA to `anatomy_unique_prefixes` splits 67 members out of existing cliques and drops
+  two obsolete EMAPA identifiers that `main` was carrying via stale UBERON xrefs.
```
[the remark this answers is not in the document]
gaurav: Outdated — that sentence is gone. The README now says restricting EMAPA "withdrew three already-published identifiers", and `clique-diff.md` explains that the three fail in two different ways (`EMAPA:35358`/`EMAPA:35459` are properly deprecated; `EMAPA:16271` has no triples at all). EMAPA also ended up *not* restricted, so nothing is withdrawn in the shipped configuration.

src/ubergraph.py
```diff
@@ -590,37 +631,62 @@ def _assert_upper_case_prefixes(argument_name, prefixes):
...
     types2relations = {"xref": "xref", "exact": "oio:exactMatch", "close": "oio:closeMatch"}
     if set_type not in types2relations:
         return
+    if hierarchy_predicate != HIERARCHY_SUBCLASS_OF and set_type != "xref":
+        raise ValueError(
+            f"hierarchy_predicate={hierarchy_predicate!r} is only supported for "
+            f"set_type='xref'; set_type={set_type!r} hardcodes rdfs:subClassOf. "
+            "Extend get_subclasses_and_exacts() / get_subclasses_and_close() "
+            "before using a custom hierarchy predicate with those set types."
+        )
```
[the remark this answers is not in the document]
gaurav: Fixed in cac6873e. Agreed — a comment saying "never pass untrusted input" is exactly the unenforced invariant this repo's AGENTS.md warns about, and `build_sets()` already asserts `ignore_list` is upper-case rather than trusting its docstring.

`_assert_known_hierarchy_predicate()` rejects anything outside `HIERARCHY_PREDICATES`. Called at `build_sets()`'s argument boundary *and* at both interpolation sites (`get_subclasses_of`, `get_subclasses_and_xrefs`), so direct callers of those methods are covered too — writing the test surfaced that a stubbed `UberGraph` bypassed a guard placed only at the query methods.

src/babel_utils.py
```diff
@@ -1238,6 +1238,37 @@ def read_identifier_file(infile):
...
+        for lineno, line in enumerate(inf, 1):
+            if line.startswith("#"):
+                continue
+            stripped = line.strip()
+            if not stripped:
+                continue
+            x = stripped.split(" ")
+            if len(x) != 2:
+                raise ValueError(f"{fn}:{lineno}: expected two space-separated CURIEs, got {len(x)}: {line.rstrip()!r}")
+            morebad.add((x[0], x[1]))
```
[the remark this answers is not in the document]
gaurav: Fixed in f067385f — took the suggested shape: reject tabs by name, then `split()` so runs of spaces collapse. You're right that the old check failed a build over `A␣␣B`, which is unambiguous.

Kept the raise rather than reverting to a skip: a silently-ignored entry means the bad xref reappears in the compendia with nothing saying why. Tests cover both the tab rejection and the repeated-space tolerance.

docs/sources/EMAPA/filtering.md
```diff
@@ -0,0 +1,43 @@
...
+(plus the few `is_a` links) of each root.
+
+### Biolink Model registration caveat
+
+`EMAPA` is in the Biolink Model's `id_prefixes` list for `AnatomicalEntity` but **not** for
+`GrossAnatomicalStructure`. `NodeFactory.create_node()` drops identifiers whose prefix is not
+permitted for a clique's biolink class, so EMAPA terms typed as gross are not written to the
+compendium until `EMAPA` is added to `GrossAnatomicalStructure`'s `id_prefixes`. The
+source-impact report flags these with `would_be_added = false` /
+`needs_biolink_registration = true`; see `docs/AddingNewSources.md`.
```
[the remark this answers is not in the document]
gaurav: Fixed in a7bd6924. Confirmed against `get_biolink_model_toolkit("4.4.3")`: EMAPA is registered for both `AnatomicalEntity` and `GrossAnatomicalStructure`, and all 8,078 identifiers reach a compendium (2,787 + 5,291). This was the third copy of that stale claim — the others were already corrected in `write_emapa_ids()`, `AGENTS.md` and `AddingNewSources.md`.

Rewrote the section around the real caveat instead of deleting it: EMAPA is *not* registered for `Cell` or `CellularComponent`, which is exactly how `EMAPA:18428` and `EMAPA:16112` were being silently dropped until `anatomy_badxrefs.txt` broke the two bad merges.

docs/sources/EMAPA/impact-report/new-cliques.csv
SkyeAv, line 1: Do we want to make this a new standard for source docs? IK this isn't included for other sources? Maybe gitignore?

SkyeAv: I would extend this to the xrefs. It's hard to render the diffs in GitHub and I imagine some sources having an impact report that would be far too large for GitHub. I'm leaning towards not tracking these going forward.

docs/sources/EMAPA/on-addition/clique-diff.summary.json
SkyeAv, line 1: We should 100% add this to all new changes going forward. I love the summary.

input_data/anatomy_badxrefs.txt
SkyeAv, line 1: I would pick a more structured file type for this as a space delimited text file has limitations. I'm fine with it staying with this current convention but I'd prefer a csv/tsv/ndjson with a header (if relevant) or some file format that's more explicit and less up to interpretation. IK the comment does this so this is just a presonal preference thing.

src/babel_utils.py
SkyeAv, line 1: Neat idea. Goes hand in hand with the manual compendia in https://github.com/NCATSTranslator/Babel/pull/887. I could see these two's code being merged into one more general object for concision.

SkyeAv (commented):
### Summary

**Good overall PR.** Love some ideas like the new impact reports and bad xrefs. I think before we merge it's worth looking into some of the architectural patterns and standards this sets because I'm not sure all of them yet scale to all sources (i.e., the new-cliques/xref files).
- see comments for specifics

### Recommendation

Merge with minor revisions.

docs/sources/EMAPA/impact-report/new-cliques.csv
gaurav: That is true. How about if we restricted this file to 100 random rows, or the 100 longest lines? That would give SMEs something to review and would maintain some kind of record of the shape of the rows this source is adding, without taking up *too* much space. I think that's what we should for now, but if a larger file is ever useful, we could upload that to a cloud spreadsheet app (e.g. Google Drive, which is what RENCI mostly uses) and include a read-only link in the PR for SMEs to review.

gaurav: How's this for an idea for new-xrefs.tsv: instead of listing every xref, we instead group this by prefix-prefix counts, with 1-3 examples? Something like the following, but in CSV instead of JSON so it's easier for SMEs to review:

```json
{
  "prefix1": "UBERON",
  "predicate": "xref",
  "prefix2": "EMAPA",
  "count": 1454,
  "examples": [
    { "curie1": "UBERON:0000002", "label1": "uterine cervix", "curie2": "EMAPA:29927", "label2": "cervix uteri" },
    { "curie1": "UBERON:0000003" ,"label1": "naris", "curie2": "EMAPA:17847", "label2": "naris" },
    { "curie1": "UBERON:0000004", "label1": "nose", "curie2": "EMAPA:16542", "label2": "nose" }
  ]
}
```

That shows SMEs *everything* that is in that file, but is much, much more compact?

docs/sources/EMAPA/on-addition/clique-diff.summary.json
gaurav: It's pretty great! I'm not sure whether it will make sense to run this diff on the HPC cluster every time we add a new source (which is why I think the impact report is much more feasible), but it is an excellent way to see what's going on.

input_data/anatomy_badxrefs.txt
gaurav: No, you're right, that would be better. There are also a bunch of formats that do similar things (badxrefs, manual concords) that we should be able to standardize to a single format. I think YAML would be the best format IMO because that's the only one of those that includes full comment support. I've opened issue #1000 to address this: https://github.com/NCATSTranslator/Babel/issues/1000

src/babel_utils.py
gaurav: Yup, absolutely! This is actually not new code -- we've had several independent pieces of code reading badxref formats, this just centralizes it. Which will hopefully also make it easier to merge them all into one format (#1000).

PR #831, opened 2026-06-05
Download and ingest all ComplexPortal species into the MacromolecularComplex compendium (#751)

### LINKED ISSUE #751: Add additional MacromolecularComplex IDs
We currently [only include](https://github.com/NCATSTranslator/Babel/blob/704da0d1324d008e8922d77afc35dd4788b88dbe/src/datahandlers/complexportal.py#L6-L7) identifiers for *Saccharomyces cerevisiae*. Should we include others?

hyi:
The original ComplexPortal ingest pulled a single hard-coded species file and produced only
labels/synonyms for the `MacromolecularComplex` compendium. This PR downloads
**all** ComplexPortal species files, extends extraction to taxa, descriptions, and IDs, and builds
out a full unit + network + pipeline test suite around the handler. The remaining commits are
iterative hardening, simplification, and documentation worked through with Copilot and Claude. Closes #751.

## Ingest changes (`src/datahandlers/complexportal.py`)

- **Download all species files** — discover every ComplexTAB `.tsv` by scraping the EBI Apache
  autoindex listing (`fetch_complexportal_tsv_filenames`, renamed from `get_` to make the network
  call explicit) and download each over HTTPS. Raises if the listing yields zero `.tsv` files.
- **Manifest as the download sentinel** — `pull_complexportal()` writes the manifest of downloaded
  files as its last action; the Snakemake rule tracks the manifest (an earlier `download_done`
  flag was added then removed as redundant) as its output, cleanly separating the download phase
  from extraction. A missing TSV listed in the manifest raises a clear `RuntimeError`.
- **Extract taxa** — read column 3 (`Taxonomy identifier`) and write `CURIE → NCBITaxon:NNNN`,
  deduplicated by `(identifier, taxon)` so a complex conserved across species keeps all its taxa;
  rows with `-` are skipped; a defensive guard rejects an already-`NCBITaxon:`-prefixed value.
- **Extract descriptions** — read column 9 and write `CURIE → text`, deduplicated by
  `(identifier, description)` so identical text is written once but distinct descriptions
  accumulate (matching `DescriptionFactory`).
- **Generate the IDs file directly** — emit `CURIE → biolink:MacromolecularComplex` from the
  source rows so accessions with an empty recommended name are still captured, replacing the old
  `awk`-over-labels Snakemake rule.
- **Correct label dedup** — key labels on the identifier alone (first-seen wins) so the same
  accession appearing in two species files with different names can't produce a malformed labels
  file.
- **Format documentation in source** — the 19-column ComplexTAB layout now lives in
  `COMPLEXTAB_COLUMNS`/`COMPLEXTAB_HEADER` next to the code that reads it, annotating which columns
  Babel reads and which are future candidates. Plus docstrings and assorted simplification.

## Snakemake (`datacollect.snakefile`, `macromolecular_complex.snakefile`)

- New `taxa`, `descriptions`, and `ids/ComplexPortal` outputs wired through
  `make_labels_synonyms_and_taxa`; `taxafile`/`descfile` declared as explicit inputs to the
  compendium rule so the DAG reflects what `TaxonFactory`/`DescriptionFactory` read.
- Removed the redundant `macromolecular_complex_ids` awk rule; snakefmt cleanup.

## Tests

- **`tests/datahandlers/test_complexportal.py`** (unit + one network) — listing parsing, download +
  manifest, and the cross-species dedup behaviour of every output, with fixture rows built from the
  real 19-column header imported from the source module.
- **`tests/pipeline/test_complexportal.py`** (pipeline) — downloads all files and validates each
  output, including a **header-column assertion** that catches upstream format drift and a
  **cross-file consistency** check that every source `(CURIE, taxon)` pair survives into the output.
- Promoted `assert_taxa_file_valid` and `assert_descriptions_file_valid` into the shared
  `tests/conftest.py` alongside the existing `assert_*_file_valid` helpers.

## Documentation

- New `docs/sources/COMPLEXPORTAL/Ingestion.md` and `docs/sources/DownloadPatterns.md` (HTTP
  autoindex listing vs FTP `NLST`), indexed from `docs/sources/README.md`.
- `tests/README.md` updated with the new test files, the shared assert helpers, and the
  header-assertion / cross-file techniques.
- `CLAUDE.md` updated: new assert helpers, ComplexPortal in the per-source index, and two new
  conventions (pin external column layouts in source + assert headers in tests; manifest/sentinel
  as the Snakemake output of a multi-file download).

### DIFF
--- a/src/datahandlers/complexportal.py
+++ b/src/datahandlers/complexportal.py
@@ -75,20 +75,21 @@ def _read_manifest(manifest_file):
         return [line.strip() for line in manifest if line.strip()]
 
 
-def make_labels_and_synonyms(manifest_file, download_dir, labelfile, synfile, metadata_yaml):
+def make_labels_synonyms_and_taxa(manifest_file, download_dir, labelfile, synfile, taxafile, metadata_yaml):
     filenames = _read_manifest(manifest_file)
     used_labels = set()
     used_synonyms = set()
+    used_taxa = set()
 
-    with open(labelfile, "w") as outl, open(synfile, "w") as outsyn:
+    with open(labelfile, "w") as outl, open(synfile, "w") as outsyn, open(taxafile, "w") as outt:
         for filename in filenames:
             infile = os.path.join(download_dir, filename)
             with open(infile) as inf:
                 next(inf)  # skip header
                 for line in inf:
-                    sline = line.split("\t", 3)
-                    if len(sline) < 3:
-                        raise ValueError(f"Expected at least 3 columns in {infile}, found {len(sline)}")
+                    sline = line.split("\t", 4)
+                    if len(sline) < 4:
+                        raise ValueError(f"Expected at least 4 columns in {infile}, found {len(sline)}")
 
                     identifier = f"{COMPLEXPORTAL}:{sline[0]}"
                     label = sline[1]  # recommended name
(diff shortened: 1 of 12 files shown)


src/datahandlers/complexportal.py
```diff
@@ -91,10 +91,9 @@ def make_labels_and_synonyms(manifest_file, labelfile, synfile, metadata_yaml):
 
                     identifier = f"{COMPLEXPORTAL}:{sline[0]}"
                     label = sline[1]  # recommended name
-                    label_row = (identifier, label)
-                    if label_row not in used_labels:
+                    if identifier not in used_labels:
                         outl.write(f"{identifier}\t{label}\n")
-                        used_labels.add(label_row)
```
hyi: @gaurav I understand now and agreed using identifier alone as the deduplication key is the right approach. That said, with this change, I think it is better to rename ```used_labels``` to ```used_identifiers``` instead.

gaurav (approved):
This is great, Hong! I added a few additional data extractions (descriptions and taxa) and got Claude to write some tests and some documentation for adding ingests in the future. All very simple on top of your great foundation. Thank you!

PR #853, opened 2026-06-16
fix issue #744 by handling malformed NCBIGene source alias field

### LINKED ISSUE #744: NCBIGene:828367 includes "''cytochrome P450" as a synonym
https://name-resolution-sri.renci.org/synonyms?preferred_curies=NCBIGene%3A828367

{{BabelTest|ShouldNotHaveSynonym|''cytochrome P450}}

### LINKED ISSUE #932: NCBIGene: drop the junk comma-pieces left behind by NCBI's '' quoting
Follow-up from #744. **Fixed in #853** (code) with the review artifacts in #917; will close when #853 merges.

> **Note:** this issue originally proposed *reconstructing* the quoted value by rejoining the
> fragments between the `''` markers. That turned out to be impossible — see "Why reconstruction
> doesn't work". It was rewritten around the actual bug, which is now fixed.

## What NCBI does

Some aliases contain commas. NCBI wraps those in `''…''` and then turns their **internal commas
into pipes** — the same character that delimits the `Synonyms` column, shredding the value across
it. Gene 828367 (`CYP706A2`):

```text
Synonyms[4]  : ''cytochrome P450|T12H17.100|T12H17_100|cytochrome P450|
               family 706|polypeptide 2|polypeptide 2''|subfamily A
Full_name[11]: cytochrome P450, family 706, subfamily A, polypeptide 2
```

## The bug

The #744 fix dropped the two fragments carrying `''` markers, so nothing malformed was emitted. But
the value's **middle** pieces carry no `''` at all and are indistinguishable from real aliases by
shape — so they were emitted as standalone synonyms. For `CYP706A2`: `cytochrome P450`,
`family 706`, `polypeptide 2`, `subfamily A`.

Measured over the full `gene_info.gz`: **912 junk synonyms across 276 rows, 186 distinct tokens.**
Several were actively hazardous rather than merely useless — `MET`, `CYS` and `PRO` are real human
gene symbols, and the set also included a bare `3` and mid-name slices like
`4)-dienoyl-CoA isomerase 1`.

## Why reconstruction doesn't work

You cannot rebuild the quoted value from this column. 276 rows carry an open marker; 265 of those
also have a close marker and a `Full_name` to check against, and **zero** of them rejoin to it.
(Computed by `docs/sources/NCBIGene/quoting/double_prime_report.py`.)

- **The pieces are not contiguous.** `T12H17.100` and `T12H17_100` are genuine, unrelated aliases
  sitting *inside* the span. Rejoining everything between the markers yields
  `cytochrome P450, T12H17.100, T12H17_100, cytochrome P450, family 706, polypeptide 2` — garbage.
- **The order is meaningless.** The list looks like set-iteration order; the pieces are scattered.
- **The commas became pipes**, so a pipe separating two aliases is indistinguishable from a pipe
  that was a comma inside one alias.
- **Both a quoted and an unquoted copy** of the first and last piece are present.

## The fix

Reconstruction was never needed: `Full_name_from_nomenclature_authority` / `Other_designations`
carry the value intact and correctly comma-formatted, and Babel already reads both — so the value
reaches the synonyms regardless. That also makes the junk removable, because the intact value's
comma-pieces are *exactly* the fragments to delete.

`split_ncbigene_synonym_field` now takes that value as `quoted_value` and drops any fragment
matching one of its pieces — but **only when the field carries a `''` open marker**, since absent
one nothing was shredded into it and every fragment is a real alias.

Verified against the full file:

```text
rows with a shredded value          276
synonyms emitted before the fix   1,501
synonyms emitted after the fix      589
junk synonyms removed               912
  distinct tokens                   186
```

The output is always a subset of the previous one (the fix only ever drops), and no shredded piece
survives.

## Was a genuine alias dropped? No

The obvious risk was dropping a real alias that happened to coincide with a piece of the full name.
It does not materialise: all 186 dropped tokens are mid-name fragments (`subfamily A`, `delta(2`,
`5)`), and not one is a plausible standalone alias. Every removed synonym is listed per-gene in
`docs/sources/NCBIGene/quoting/shredded_pieces.csv` (#917), generated by `shredded_pieces_report.py`,
which calls the production function twice — with and without the intact value — so the report cannot
drift from what Babel emits.

## Residual risk

The fix trusts `Full_name` to be the value that was shredded. That held for all 276 rows in the
current snapshot, but it is an inference about NCBI's behavior, not something they document. If a
future `gene_info.gz` shreds a value that is *not* in `Full_name`, its middle pieces would return.
`tests/pipeline/test_ncbigene.py` would catch that for gene 828367, but not in general.

## Tests

The two assertions that pinned the old behavior were inverted (not deleted), a new unit test covers
the drop, and the `pipeline` test asserts over the real file that gene 828367's middle pieces are
gone while the intact value survives.

hyi:
The source of this issue comes from the malformed aliases as shown from <https://eutils.ncbi.nlm.nih.gov/entrez/eutils/esummary.fcgi?db=gene&id=828367&retmode=json>.

Follow-on work by @gaurav: further investigation (#917, which characterizes all 70.5M rows of `gene_info.gz`) found that `''` means two different things depending on context:

1. **Sometimes `''` really is quoting.** NCBI wraps a comma-containing alias in `''…''` and turns its internal commas into `|` — the column's own delimiter. We can't reconstruct the original value, because NCBI also reorders the fragments and interleaves them with unrelated aliases (0 of the 265 checkable rows rejoin cleanly). We don't need to: the correct value is carried intact by `Full_name_from_nomenclature_authority` / `Other_designations`, which Babel already reads.

2. **Sometimes `''` is a double-prime in a genuine name** — `U2B''`, RNA polymerase subunit `beta''`, `6''-O-malonyltransferase`. These are 25,335 occurrences versus 276 shredded values, so we can't drop every synonym containing `''`; the original fix was discarding ~4,000 real synonyms.

What distinguishes them is *position*: a **leading** `''` never begins a genuine name. So `split_ncbigene_synonym_field` drops a fragment only when it is a leading open marker, or a trailing close marker **in a field that also contains an open marker**. A trailing `''` on its own is kept verbatim, and a self-contained `''X''` is unwrapped to `X`.

That still left the shredded value's **middle** pieces (#932) — `family 706`, `subfamily A`, `MET`, even a bare `3` — which carry no `''` at all and look exactly like real aliases. Over the full file that was **912 junk synonyms across 276 rows**; `MET` and `CYS` are real human gene symbols, so this was a name-resolution hazard, not just noise. The same intact `Full_name` value fixes it: its comma-pieces are exactly the fragments to remove.

Verified against the full file: 912 junk synonyms removed, none left, and the output is always a subset of the previous one (this only ever drops). Fixes #744, fixes #932.


gaurav (approved):
Thanks for the fix, Hong! I dug into this in more detail in https://github.com/NCATSTranslator/Babel/pull/931 and found a whole ton of edge cases, but they were pretty easy to fix on top of the fixes you'd already written.

PR #869, opened 2026-06-20
Right-size default SLURM resources from the full babel-1.17 run

gaurav:
Right-sizes the cluster-wide SLURM defaults using a **full `babel-1.17` run** (345 rules
benchmarked), analyzed with `babel-slurm-resources`:

- Lowers `slurm/config.yaml` defaults from `mem: 64G` / `cpus_per_task: 4` to
  `mem: 16G` / `cpus_per_task: 1`. **90% of rules (312/345) peaked below 8 GB** and all but
  the DuckDB export rules used a single core, so the old 64 GB default packed only ~3 light
  jobs onto a 191 GB batch node. The run wasted ~18 TB of reservation (requested minus used)
  across rules with a known request.
- Adds explicit `mem` overrides to the **six rules that run on the bare default and peak above
  16 GB** — without a block they would OOM-kill under the new default:

  | rule | peak RSS | new `mem` |
  |---|---|---|
  | `drugchemical_conflation` | ~57G | 64G |
  | `geneprotein_conflation` | ~48G | 64G |
  | `get_uniprotkb_labels` | ~40G | 48G |
  | `hmdb_labels_and_synonyms` | ~30G | 48G |
  | `check_protein_completeness` | ~21G | 24G |
  | `get_chemical_unichem_relationships` | ~21G | 24G |

- Gives the two generic DuckDB export rules (`export_compendia_to_duckdb`,
  `export_synonyms_to_duckdb`) `cpus_per_task=4`. DuckDB auto-threads and used up to ~2.6
  cores; with the new `cpus_per_task=1` default the slurm executor would otherwise pin them to
  one core (the `resources.cpus_per_task` value takes precedence over `threads:`). Packing for
  these is memory-bound (512G/128G), so the cores are effectively free.

Rules that already carry explicit large `resources:` blocks (the big compendia, the largemem
DuckDB report rules, etc.) are untouched.

Override sizes are biased **down** toward the observed peak (~10–15% headroom, smallest sensible
bucket) rather than a padded safety factor: an OOM is cheap to spot (we track per-rule peak RSS)
and re-run at a higher limit, whereas padding every job slowly ratchets the whole cluster upward —
which is exactly the over-provisioning the old 64 GB default represents.


gaurav:
Heads-up while right-sizing resources (also added as a TODO in the PR description):

Dropping the default `cpus_per_task` from 4 to 1 under-provisions any rule that runs DuckDB with `threads > 1`. On current `main` the only such DuckDB report rule is **`check_for_duplicate_clique_leaders`** — it runs `threads=4` in `duckdb.snakefile` but sets no explicit `cpus_per_task`, so today it relies on the default of 4. Under the new default it would get 1 core for 4 threads and needs an explicit `cpus_per_task=4`.

The other report rules (`check_for_identically_labeled_cliques`, `check_for_duplicate_curies`, `generate_curie_report`, `generate_clique_leader_report`) all run `threads=1`, so the new default covers them.

Re-running the resource calc against the full babel-1.17 HPC run should catch this from actual core usage; flagging it as a backstop. Surfaced while closing #703 (its two-pass DuckDB work is already on `main`; this `cpus_per_task`/`threads` coupling was the one bit not yet covered).

gaurav:
**Correction / resolved** (re: the `cpus_per_task`/`threads` note above):

On a closer look, the premise was off. The `"threads": 4` in `check_for_duplicate_clique_leaders`
is a **DuckDB config pragma passed into the report call**, not a Snakemake `threads:` rule
directive — Snakemake never sees it, so it never drove `cpus_per_task`. The full babel-1.17
benchmark confirms the rule used only ~1.4 cores on a 166s job, so it needs no `cpus_per_task`
override under the new `cpus_per_task=1` default.

The real (smaller) version of the concern is the generic DuckDB **export** rules
(`export_compendia_to_duckdb`, `export_synonyms_to_duckdb`), which let DuckDB auto-thread and used
up to ~2.6 cores. Those now carry an explicit `cpus_per_task=4` — and it has to be
`cpus_per_task`, not `threads:`, because the slurm executor takes `resources.cpus_per_task` in
precedence over `threads:`, and the new default populates that key on every job.

Nothing to do here beyond what's in the current diff; flagging so the earlier note isn't read as
an outstanding item.

Issue #827, opened 2026-06-04
Preferred name incorrect: DRUGBANK:DB10626 "Trout" should be "Trout allergenic extract"


gaurav:
### CURIE

DRUGBANK:DB10626

### Current preferred name

Trout

### Correct preferred name

Trout allergenic extract

### NodeNorm instance or URL

https://nodenormalization-sri.renci.org/1.5/get_normalized_nodes?curie=DRUGBANK%3ADB10626&conflate=true&drug_chemical_conflate=true&description=false&individual_types=false&include_taxa=true

### NodeNorm JSON response (optional)

```json
{"DRUGBANK:DB10626":{"id":{"identifier":"DRUGBANK:DB10626","label":"Trout"},"equivalent_identifiers":[{"identifier":"DRUGBANK:DB10626","label":"Trout"},{"identifier":"UMLS:C2725895","label":"trout allergenic extract"},{"identifier":"RXCUI:882482"}],"type":["biolink:ChemicalEntity","biolink:PhysicalEssence","biolink:ChemicalOrDrugOrTreatment","biolink:ChemicalEntityOrGeneOrGeneProduct","biolink:ChemicalEntityOrProteinOrPolypeptide","biolink:NamedThing","biolink:PhysicalEssenceOrOccurrent"]}}
```

### Also tracked at

https://github.com/NCATSTranslator/Feedback/issues/1284

### BabelTest entry (optional)

```text
{{BabelTest|HasLabel|DRUGBANK:DB10626|Trout allergenic extract}}
```

PR #872, opened 2026-06-23
fixed issue #827 with DRUGBANK:DB10626 preferred name

hyi:
@gaurav This issue is triggered by upstream DRUGBANK source having a broad Trout as the preferred label for this curie. Even though both UMLS C2725895 and [rxcui:882582](https://rxnav.nlm.nih.gov/REST/rxcui/882482/properties.json) have the correct preferred label, DRUGBANK preferred label is chosen due to the preferred_name_boost_prefixes configuration for biolink:ChemicalEntity in config.yaml file that includes DRUGBANK in the list. Since this is just one example where boosted DRUGBANK label does not give us the correct preferred name, I fixed this issue by adding a configurable label_overrides.yaml which will be loaded to correct upstream DRUGBANK preferred label for this curie to the correct label. Let me know if there is a better approach to fix this issue, though.

### DIFF
--- a/input_data/label_overrides.yaml
+++ b/input_data/label_overrides.yaml
@@ -1,0 +1,22 @@
+# Manual CURIE-specific label corrections.
+#
+# Use this file when an upstream source emits a primary label that is too broad,
+# misleading, or otherwise unsuitable for Babel output, and the correction should
+# apply only to that exact CURIE.
+#
+# Required fields:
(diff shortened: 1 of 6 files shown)


gaurav:
Thanks, Hong! I don't like the idea of special-casing the label for this particular concept, since I think we can fix in other ways, but I think having a mechanism for doing this in general would be good to have. I'll think about a bit more next week.

PR #876, opened 2026-06-26
Add bounded retry/backoff to TripleStore SPARQL execution

gaurav:
## Summary

Hardens `TripleStore.execute_query()` against transient SPARQL endpoint failures with a bounded exponential-backoff retry.

- Retries on HTTP 429 (rate limiting), HTTP 5xx, `JSONDecodeError`, and `OSError` (covers `TimeoutError`, `URLError`, connection resets, etc.); never retries other 4xx (deterministic client errors).
- Exponential backoff: attempt 1 → 1 s, attempt 2 → 2 s, attempt 3 → 4 s, …
- Attempt count and base delay come from a new `sparql:` block in `config.yaml` (`max_attempts`, `retry_base_delay_seconds`), overridable per call.
- Input validation raises `ValueError` on `max_attempts < 1` or `retry_base_delay_seconds < 0` before any retry logic runs.
- Internal helpers (`_dispatch_with_retries`, `_wait_before_retry`) are private.
- Module logger lowered from `ERROR` to `WARNING` so retry/backoff messages are visible.

The `uv.lock` babel-pipeline version line is synced `1.14` → `1.17` to match `pyproject.toml` (the lock was already stale on `main`).

## Test plan

- [x] `uv run pytest -m unit -q tests/test_triplestore.py` — retries on JSON-decode
  error, HTTP 503, and HTTP 429; does not retry HTTP 400; raises after all attempts
  exhausted; rejects invalid `max_attempts` and negative delay.
- [x] `uv run ruff check`

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -28,6 +28,15 @@ tmp_directory: babel_downloads/tmp
 # HTTP settings.
 http: {}
 
+# SPARQL query settings.
+sparql:
+  # How many times to attempt a SPARQL query before giving up. The first attempt counts,
+  # so max_retries=3 means one initial try plus two retries.
+  max_retries: 3
+  # Base delay in seconds between retries. Actual delay grows exponentially with each attempt
+  # (attempt 1 → 1s, attempt 2 → 2s, attempt 3 → 4s, …).
+  retry_base_delay_seconds: 1
+
 # Maps Python compendium names (as used in src/createcompendia/ and tests) to the
 # directory names that Snakemake uses under intermediate_directory.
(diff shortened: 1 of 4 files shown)


gaurav:
Fixed in d7b88f54. Added `ValueError` guards for `max_attempts < 1` and `retry_base_delay_seconds < 0` at the top of `execute_query()`, so bad config values fail loudly before any retry logic runs. Also renamed `max_retries` → `max_attempts` throughout (config + code + tests) to make the semantics unambiguous, and added tests for both invalid-config cases.

PR #883, opened 2026-06-29
Activate the MONDO close-match guard in disease glom (with build-vs-build clique-diff tool)

gaurav:
> [!IMPORTANT]
> **Superseded by #888.** The MONDO close-match guard fix has been split onto a clean,
> `main`-based branch in #888 (the same `x[1]`→`x[2]` change applied directly to
> `build_compendium`, plus a focused regression test), and the build-vs-build clique-diff tool
> already merged separately via #885. If #888 merges first, this guard change drops out of this
> PR automatically once `main` is merged in — leaving #883 with nothing unique to contribute.
> Prefer reviewing #888 for the guard fix in isolation; this PR is kept only as the home of the
> full add-mpo-base before/after impact analysis under
> `docs/pipelines/diseasephenotype/mondo-close-match-guard/`.

## Summary

Fixes a latent bug in how the disease/phenotype build reads the `MONDO_close` concord, which
left glom's close-match guard a **silent no-op**, and adds a reusable build-vs-build clique-diff
tool plus the before/after impact analysis that this change requires.

This is the dedicated follow-up promised in #790, where the same code was touched only enough to
stop the build crashing (validating the real 3-column `MONDO_close` format while *preserving*
the dormant `x[1]` behaviour). Activating the guard changes disease clique merging broadly and is
orthogonal to adding MP, so it lives here with its own evidence.

> [!NOTE]
> Based on `add-mpo` (#790), not `main` — it builds on that PR's `compute_cliques_for_impact_report`
> refactor. Merge #790 first; this PR will retarget to `main` automatically.

## The bug

`compute_cliques_for_impact_report()` read `MONDO_close` (a 3-column `subject  predicate  object`
concord) with `close_mondos[x[0]].add(x[1])` — keying the MONDO subject to **column 2, the
predicate** (`oio:closeMatch`) instead of **column 3, the close-match object**. No clique ever
contains the literal string `oio:closeMatch`, so glom's `close=` guard (`if cd in newset`) never
matched. The guard has therefore never fired, on `main` and before it; close (but not exact)
matches were free to collapse into exact cliques.

The fix is one line of intent: key on `x[2]`, matching the `(stuff[0], stuff[2])` parsing every
other concord in that function already uses.

## Impact (why this needs SME review)

Measured on a full local `disease` build (Biolink 4.4.3), same intermediate ids/concords,
differing only in `x[1]` vs `x[2]`. glom is deterministic — two `x[1]` builds were clique-identical
— so every difference is signal, not run-to-run noise.

- **1,219 identifiers, all MEDDRA, are dropped from `Disease.txt`.** They were present only because
  the dormant guard let them merge into a MONDO disease clique; correctly kept out, they have no
  independent disease typing in this pipeline and leave the compendium entirely.
- **1,191 of 365,465 Disease cliques** lose ≥1 member; net Disease clique count +20, PhenotypicFeature +13.
- `PhenotypicFeature.txt` is essentially unaffected (the close pairs are MONDO-subject).

Dropping 1,219 MEDDRA codes is *correct* if they are genuinely distinct concepts (the guard's
premise), but a **coverage loss** if anything relies on that MEDDRA → MONDO normalization. Per the
project rule against dropping valid identifiers without good reason, this is an SME call — the
evidence is in `docs/pipelines/diseasephenotype/mondo-close-match-guard/` (`README.md`,
`clique-diff.csv`, `dropped-members.csv`, `summary.json`).

## New tool: `babel-clique-diff`

A reusable build-vs-build clique diff (`tools/clique_diff`) that compares the finished compendia of
two builds and reports split/merged/lost cliques and, crucially, dropped CURIEs.

It is deliberately **distinct from `source-impact-report`**: that answers "what does adding *source
X* do?" by re-glomming intermediates with vs. without one source; this answers "how did the cliques
change between *build A* and *build B*?" given the same inputs but different code/config/data.
Because it works on finished compendia rather than glom state, it can also compare a local build
against a published `stars.renci.org` build with no re-glom — useful for validating any glom-logic
change (close-match, `unique_prefixes`, overuse filtering) or as a release regression check.

## Test plan

- `uv run ruff check` / `ruff format --check` — clean.
- `uv run rumdl check .` — clean.
- `uv run pytest -m unit -q` — green (243 passed), including new `tests/tools/test_clique_diff_tool.py`.

## Out of scope

- The MP source-impact report on #790 reflects `x[1]` (the production behaviour of that PR) and is
  not regenerated here.
- Deciding whether to keep the guard active — that is the SME question this PR's evidence exists to inform.


gaurav:
Heads up: the `babel-clique-diff` tool commit from this PR has been extracted into its own standalone PR off `main` (#885) so it can land independently of the MONDO close-match-guard change here. Once #885 merges, please rebase this PR to drop its now-duplicated `babel-clique-diff` commit. The new "MP kept disjoint from HP" PR (#886) depends on #885.

gaurav:
Closing: everything in this PR now lives either in `main` or in #888.

**Verified against `main` (9058e634) and #888 (`fix-mondo-close-guard`):**

- **`babel-clique-diff` tool** (`tools/clique_diff/`, `pyproject.toml` entry point, `docs/tools/README.md`, `tests/tools/test_clique_diff_tool.py`) — merged via #885, in a better shape than here: the diff logic now lives in `src/model/compendium_diff.py` with `src/tools/clique_diff/cli.py` as a thin CLI, and the tests moved to `tests/tools/clique_diff/test_cli.py`.
- **The MP base** this branch was stacked on (#790, since closed) — the MP source, its docs under `docs/sources/MP/`, and `tests/pipeline/test_mp_pipeline.py` are all in `main`.
- **The guard fix itself** (`close_mondos[x[0]].add(x[1])` → `x[2]`) — carried by #888, applied directly to `build_compendium` on a clean `main` base, with 164 lines of new concord-reader symmetry tests in `tests/createcompendia/test_diseasephenotype.py` that this PR did not have.
- **The before/after impact analysis** (`docs/pipelines/diseasephenotype/mondo-close-match-guard/`) — superseded by `docs/sources/MONDO/mondo-close-match-guard/` on #888, which is a strict superset: same clique diff (1,218 dropped MEDDRA identifiers, matching the ~1,219 reported here), plus the `split-cliques.csv` analysis of the 37 cliques that structurally split — the changes that actually need SME review — and the explanation of why `glom()` ejects a legitimate non-MEDDRA identifier as collateral damage when the guard fires.

The one file with no direct counterpart on #888 is `dropped-members.csv`, but it is redundant: all 1,218 dropped CURIEs are fully enumerated (no truncation) in the `example_members` column of #888’s `clique-diff.csv`, keyed to the MONDO clique each left.

Nothing needs cherry-picking. The guard fix continues on #888.

PR #975, opened 2026-07-27
chore: swap build backend from setuptools to maturin

SkyeAv:
Swaps the build backend from setuptools to maturin so Babel builds as a mixed Rust/Python
package, ahead of landing native (Rust) functionality. This is the backend swap only — the Rust
side is an inert placeholder module, the importable package stays `src`, and nothing under it
moves or renames.

> **Update (2026-08-10):** The module is no longer an inert placeholder — it now ships its first real
> accelerator: a Rust union-find that **replaces `babel_utils.glom()`**, the O(N²) clique-builder behind
> the ~5.5 h single-core `protein_compendia`/`chemical_compendia` rules. It is **Rust-only**: the
> `BABEL_DISABLE_RUST` env var and the Python fallback are removed, the Python `glom` is deleted, and the
> A/B-testing language is stripped from the docs/tests. `ABI_VERSION` is now 2 (rust-version 1.80). All
> ~12 callers are unchanged (the Rust `glom` is re-exported from `src.babel_utils`). **Offline benchmark:**
> ~103× faster than the Python `glom` on a 6,000-merge chain (12,000 pairs) with identical cliques; the
> speedup grows with N (9× → 100×), the signature of removing the O(N²). 449 unit tests pass; cargo
> fmt/clippy/test, ruff, and rumdl clean. Full details in the review comment below.

> **Update (2026-07-30):** [#987](https://github.com/NCATSTranslator/Babel/pull/987) is merged into
> this branch and supersedes some placeholder details below — the module is renamed `rs` → `_accel`
> (import via `src/accel.py`, with a Python fallback and a stale-ABI guard), Docker/CI use rustup
> pinned to Rust 1.95.0 (not apt cargo / `@stable`), and the formatting lint runs `uv run
> --no-project` so it doesn't build. See #987 for those specifics.

### Build backend
- **`pyproject.toml`:** `[build-system]` is now `requires = ["maturin>=1.0,<2.0"]` /
  `build-backend = "maturin"`; dropped `[tool.setuptools.packages.find]`.
- **`[tool.maturin]`:** `module-name = "src.rs"` places the compiled extension at `src/rs.*` and
  tells maturin to bundle the `src` package (auto-resolved to the repo root — src-layout detection
  doesn't fire because there's no `src/babel_pipeline/__init__.py`); `manifest-path =
  "rust/Cargo.toml"` keeps Cargo off the Python `src/` dir.
- **`dev` group:** added `maturin>=1.0,<2.0` so `uv run maturin develop` works locally (the
  `[build-system].requires` entry is build-isolation only).

### Rust stub
- **`rust/Cargo.toml` + `rust/src/lib.rs`:** empty `#[pymodule] fn rs` (pyo3, `abi3-py311` +
  `extension-module`), importable as `from src import rs`. Real functionality lands in a follow-up.
- **Layout:** Rust lives under `rust/` so Cargo's default `src/lib.rs` never collides with the
  Python `src/` package.

### Build prerequisites (CI + Docker)
- **`.github/workflows/test.yml`:** added `dtolnay/rust-toolchain@stable` before `uv sync` in both
  jobs — every `uv sync` now compiles via cargo, and `ubuntu-latest` ships no Rust.
- **`Dockerfile`:** `apt-get install -y cargo` (system-wide, before `USER nru`) for the same reason.
- **Accepted caveat (softer than it first reads):** a Rust toolchain is a build prerequisite, but uv
  bootstraps one itself (via `puccinialin`) when cargo is absent — a silent ~600 MB download into a
  platform cache, not a failure. The one place it genuinely fails is `uv sync --frozen` on the
  Hatteras login node (see `slurm/README.md`); CI and Docker install rustup explicitly.

### Housekeeping
- **`.gitignore`:** ignore `rust/target/` and the `maturin develop` binaries
  (`src/rs.*.so/.pyd/.dylib`).
- **`uv.lock` + `rust/Cargo.lock`:** regenerated and committed (CI runs `uv sync --frozen`).

### Behavior change (worth knowing)
- **The wheel now ships the 19 `.snakefile`s.** `main`'s setuptools wheel shipped zero non-Python
  files, so an installed `babel-pipeline` previously could not run the pipeline; maturin bundles all
  of `src/*`, which includes the `.snakefile`s. An accidental fix, but a behavior change.

### Design
- **Why maturin needs a stub:** maturin is fundamentally a Rust build tool — every layout it
  supports requires a `Cargo.toml`, so a bare backend swap with no Rust won't build. The empty
  `#[pymodule]` is the smallest thing that makes maturin viable now.
- **Why `module-name = "src.rs"`:** the import package is literally named `src` (flat layout).
  Verified against maturin's resolver (`project_layout.rs` / `module_writer/mod.rs`): this bundles
  all of `src/*` and drops the extension at `src/rs.<so>` without moving or renaming anything.
- **Deferred:** real Rust functionality; a manylinux/abi3 publishing matrix (the stub builds a
  local `cp311-abi3` wheel).

### Testing
- `uv sync` — builds `babel-pipeline` via maturin → cargo; installs `maturin==1.14.1`.
- `uv run python -c "import src; from src import rs; from src.tools.clique_diff import cli"` — ok.
- `uv run pytest -m unit -q` → `402 passed, 105 deselected`.
- `uv build` → `babel_pipeline-1.17-cp311-abi3-linux_x86_64.whl`; `unzip -l` shows `src/__init__.py`,
  `src/node.py`, `src/tools/...`, `src/rs.abi3.so`, and all 4 console scripts in `entry_points.txt`.
- `uv run babel-clique-diff --help` — resolves.

## Questions for the reviewer
- **Rust toolchain as a prerequisite.** Every build invokes cargo, but uv self-bootstraps a toolchain
  (~600 MB) when it's missing, so the practical cost is a one-time download, not a hard blocker —
  except `uv sync --frozen` on Hatteras, which does fail without a toolchain. Fine for a stub-only
  PR, or would you rather gate the maturin switch behind the PR that adds real Rust?


SkyeAv:
This would allow me to expose rust code directly to python. This backend is the base commit I need before I can start refactoring in that direction

SkyeAv:
Addressed both Copilot review points in cc0192d7:

### Reproducible Rust toolchain (`.github/workflows/test.yml`)
- Added a checked-in `rust-toolchain.toml` pinning `channel = "1.95.0"` (the toolchain used locally), so local builds, CI, and any rustup-based environment all compile the maturin/PyO3 extension with the same compiler.
- Pinned the CI action ref `dtolnay/rust-toolchain@stable` → `@1.95.0` in both test jobs, so CI no longer tracks the moving stable channel. The action ref and `rust-toolchain.toml` are cross-referenced (comment on each) to be bumped together.

### Lockfile-reproducible Docker build (`Dockerfile`)
- `uv sync` → `uv sync --frozen`, so the image installs exactly the committed `uv.lock` with no re-resolution — matching CI's `uv sync --frozen`. This matters more now that the step also compiles Rust via maturin/cargo.

**Verified locally:** `rustup show active-toolchain` reports `1.95.0-x86_64-unknown-linux-gnu (overridden by rust-toolchain.toml)`, and both `cargo check` and a full `uv run` maturin rebuild of the extension succeed under the pin.

SkyeAv:
## Status on the two branches offered into this PR

**#987 (Rust plumbing) — merged** (`b10b48a4`). Hardens the extension contract: single import point
`src/accel.py` with a Python fallback + stale-ABI guard, `BABEL_DISABLE_RUST`, the three maturin
build-gap fixes (rustup in Docker, `--no-project` lint, slurm/docs notes), and cargo fmt/clippy +
rust-cache CI. Module renamed `rs` → `_accel`. Conflicts resolved by keeping the Rust 1.95.0 pin and
adding `Swatinem/rust-cache`.

**#988 (boundary-cost experiment) — closed unmerged.** Its headline — "in-process can save at most
~4 min / 1.3%, a ceiling no pyo3 can beat" — bounds only the cost of a *single* Python↔Rust handoff,
and benchmarks a misuse of Rust (one function converted then reserialized, single-threaded). Verified
empirically: chaining heavy routines on native data skips the "unavoidable floor" (0.67 s vs 12.4 s,
−95%), and `py.allow_threads` gives near-linear multi-core scaling in-process (~10× at 16 threads vs
~1× GIL-held). Both patterns are already in production in Tablassert. Full pushback lives on #988.

**Decision:** in-process pyo3 (this PR) is the default mechanism — less code per accelerated routine
and no new file formats, `shell:` rules, or binaries; out-of-process (B/C) reserved for stages already
split at a file boundary.

**Follow-ups harvested from #988 (no Rust needed):** #996 (JSONL, not `repr(set)`, for the clique
boundary), #997 (cache `get_biolink_model_toolkit`), #998 (read `icRDF.tsv` once). Two more Rule 0
fixes are already open: #984, #985.

CI is green (ruff, snakefmt, cargo fmt/clippy, rumdl, unit-tests).

rust/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+/// Bumped by hand whenever a function's signature or semantics change, in the same commit.
+///
+/// `src/accel.py` compares this against its own `_REQUIRED_ABI_VERSION` at import time and raises
+/// if they disagree. That matters because the extension is installed editable: the compiled
+/// artifact sits in the checkout at `src/_accel.*.so` and a `git pull` does not rebuild it, so
+/// without this check a stale binary would be used silently for a 12-hour rule.
+///
+// ponytail: a hand-bumped integer. Move to a build-time hash of this crate's sources if anyone
+// forgets to bump it twice.
+const ABI_VERSION: u32 = 1;
```
gaurav: 1. Do we really need this? I don't think recompiling the Rust code for every Babel build will be a huge imposition.
2. Is there any value for storing this in config.yaml rather than here?

tests/test_accel.py
```diff
@@ -0,0 +1,120 @@
...
+
+@pytest.mark.unit
+def test_the_built_extension_matches_the_version_python_expects():
+    """If the extension is built, its ABI_VERSION must be the one this checkout expects.
+
+    Both constants are bumped by hand in the same commit. Failing in CI after a Rust change means
+    the bump was forgotten; failing locally means the checkout needs
+    `uv sync --reinstall-package babel-pipeline`.
+    """
+    if src.accel.accel is None:
```
gaurav: I don't understand -- shouldn't this be built on `uv sync`? I'm concerned that this test will start skipping at some point but we won't notice it -- a failure would be better IMO.

rust/src/lib.rs
SkyeAv: 1. It's not about build cost — it's about the editable install. The compiled `.so` sits in the checkout (`src/_accel.*.so`, gitignored), and a `git pull` that brings new Rust source rebuilds nothing (`uv sync` only rebuilds when `pyproject.toml`/`uv.lock` change). Without the guard, a stale binary silently runs 12-hour rules; with it, the run fails at DAG-parse in the first second, with the fix command in the message. Cost is one integer comparison at import, bumped in the same commit as the change that stales the binary.
2. No — the check compares what was **baked into the `.so` at compile time** against what the Python checkout expects; `config.yaml` is runtime data and can't detect a stale binary. The two halves (here and `_REQUIRED_ABI_VERSION` in `src/accel.py`) also need to move in one commit. Same reason `BABEL_DISABLE_RUST` isn't in `config.yaml`: config is threaded into Snakemake `params`/output paths, and an implementation detail shouldn't be able to perturb the DAG.

tests/test_accel.py
SkyeAv: Yes — `[tool.uv] package = true` means every `uv sync` builds the extension (uv even bootstraps a toolchain if cargo is missing), and CI always has it, so this test runs there today. But your concern is right: if the build ever silently regressed, the skip would hide it. The no-extension fallback is covered separately (`test_a_missing_extension_falls_back_instead_of_raising` fakes the absence), so I've removed the skip — the test now fails instead: 6ebd4e4b.

rust/README.md
```diff
@@ -0,0 +1,150 @@
...
+```python
+from src.accel import accel
+
+def read_something(path):
+    if accel is not None:
+        return accel.read_something(path)
+    return _read_something_python(path)
+```
+
+`accel` is the compiled module when it is present, importable and current, and `None` otherwise.
```
gaurav: Is there any case where accel is not present?

gaurav (changes requested):
Thanks so much for all of this work, Skye! Would it be possible to include an example of Rust usage in this file so I can better see how a Rust rule could be incorporated into Snakemake? If there isn't an obvious candidate to pull in from https://github.com/NCATSTranslator/Babel/pull/588, I wondered if https://github.com/NCATSTranslator/Babel/issues/1004 might be a good candidate for Rustification: there's a bunch of different places where the text validation could go, so even if it doesn't make sense to replace to *Factory classes with something better (either entirely in Python or with some Rust code), it might make sense to add a `check_encoding(output.synonyms_file)` every time we create a synonyms to check the format. What do you think? Feel free to pick another function to Rustify or we can discuss this in person if useful -- I wouldn't try rustifying `glom()` or `write_compendia()`, which are probably the top candidates for this, because they're crucial functions that don't have any (?) tests yet and so introducing bugs in them would be baaaad, or the PubMed import, which I'm writing a complete replacement for in https://github.com/TranslatorSRI/pubmed2db as per https://github.com/NCATSTranslator/Core-Components-Working-Group/issues/15. Let me know what you think! If you think this first-Rust-function would make more sense to do in a separate PR, then I'm happy to merge this for now and we can figure that out there, but I think it would be useful to put this example next to the new Rust documentation and pyproject.toml/Cargo settings we are developing here.

SkyeAv:
## First real `_accel` accelerator: `glom()` union-find (commit 53ed0670)

This replaces the Python `babel_utils.glom()` clique-builder with a Rust union-find, and makes the
extension **Rust-only**. Details below.

### Why
`glom()` is the union-find in every compendium build. The Python version is **O(clique) per merge** —
it copies the whole merged clique (`set().union(...)`), re-scans it for the `unique_prefixes` /
`KEGG` / `PUBCHEM` / `close` checks, and reassigns every member — so a clique growing to N costs
**O(N²)**. That's a large share of the ~5.5 h, single-core, 246–335 GB `protein_compendia` /
`chemical_compendia` rules.

### What changed
- **New `rust/src/glom.rs`:** disjoint-set forest over interned `u32` CURIE ids — path compression +
  union by size; constraint checks (unique-prefix counts, garbage flag, `close`) maintained
  incrementally per root → O(#unique_prefixes), not O(clique).
- **Serialization minimizers:** merged cliques grow the *larger* live Python `set` in place (only the
  smaller side's dict entries are reassigned → O(N log N) writes), and a cross-call cache keyed by the
  dict stops repeated calls from re-reading/re-interning the whole state.
- **Parallel where sound:** the initial CURIE interning runs on rayon over a sharded map. The merge
  loop stays **sequential** because rejection is order-sensitive (reordering would change the output).
- **Rust-only (per direction):** removed `BABEL_DISABLE_RUST` and the missing-extension fallback from
  `src/accel.py`; deleted the Python `glom` (−135 lines); stripped the A/B language from
  `rust/README.md`, `AGENTS.md`, `src/accel.py`, and `lib.rs`. Correctness is guarded by tests, not a
  fallback. `ABI_VERSION` 1→2; `rust-version` 1.63→1.80 (`LazyLock`).

The drop-in contract is preserved: it mutates `conc_set` in place, returns `None`, accepts 1–2 element
groups of `str`/`LabeledID`, and takes `unique_prefixes` as any iterable (a bare string is iterated
per-char, exactly as before — `test_uberon` relies on it). All ~12 callers are unchanged.

### Correctness notes
- **Rejected groups don't add new members** (faithful to Python). Callers pre-register identifiers
  first, so this only affects never-registered ids; both behaviours are pinned by tests.
- **A bug the cache exposed (fixed):** the all-new-clique branch reset the keeper's `up_count` then
  re-added `up_count[id]` for the keeper itself (now zero), silently dropping its unique-prefix
  contribution and defeating rejection. The rebuild-per-call path masked it; the cache surfaced it via
  `test_mp_included_in_unique_prefixes_blocks_same_prefix_merge`. Fixed by computing each member's
  contribution from its prefix.

### Benchmark (offline, synthetic)
`data/benchmark_glom.py` races the git-extracted Python `glom` vs the Rust one on identical inputs
(one long merge chain + noise, 8 batches on the same dict) and asserts the cliques match. Machine:
24-core AMD Ryzen AI 9 HX 370, debug-profile Rust build (conservative).

| chain | pairs | python | rust | speedup | match |
|---:|---:|---:|---:|---:|:--:|
| 1500 | 7500 | 0.416s | 0.046s | **9.1×** | True |
| 3000 | 9000 | 1.667s | 0.067s | **24.8×** | True |
| 4500 | 10500 | 4.013s | 0.056s | **71.3×** | True |
| 6000 | 12000 | 6.921s | 0.069s | **99.9×** | True |

Headline (chain=6000): **Python 7.0s → Rust 0.07s (~103×)**, identical 3,001 cliques. The speedup
**grows with N** while Rust stays near-flat — the signature of removing the O(N²). At production scale
the quadratic term dominates more, so the real win should be larger; #990 (profile on real data) should
confirm where those rules are actually bound.

### Verification (all offline — no BABEL run)
- `uv run pytest -m unit` → **449 passed** (was 437; +12 new glom semantics tests). The suite exercises
  `glom` through its real callers — the regression net now that there's no Python reference.
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` (0 warnings), `cargo test` — clean.
- `ruff check .`, `ruff format --check`, `rumdl` (0.2.50, as pinned in CI) — clean.
- `git grep BABEL_DISABLE_RUST` → none outside scratch notes.

### Questions
- The cross-call cache keys on the dict's address and rebuilds on entry-count/`unique_prefixes`
  mismatch; it assumes a dict is only mutated through `glom` (true today). OK with that invariant, or
  prefer the simpler rebuild-every-call (still correct, just re-reads state per call)?
- Benchmark is a laptop-scale cost model. Want the coarse builder that also fuses file reading
  (`read_concord_file` / `remove_overused_xrefs`) for the hottest callers in this PR, or as a follow-up?

SkyeAv:
@gaurav LMK if you want any addtional changes. I drafted this with an LLM and had it benchmark the optimizations to glom and it seems pretty promising.

SkyeAv:
@gaurav this ran successfully at `/projects/babel/runs/goetzs/RUST-GLOM-1.18` on HT1. I don't know how much faster it really was of the top of my head but I was hoping you could let me know where to check the speed. I'll say it did seem MUCH faster running it though.

SkyeAv:
Closing for now: we decided the maturin build-backend migration is too much additional complexity to take on at the moment. The in-process pyo3 approach stays as-is via setuptools + maturin-independent build; revisit if the build-backend pain points become blocking. Branch `feat/migrate-to-maturin-build-backend` preserved.

PR #988, opened 2026-07-30
EXPERIMENT (AI-written, undecided): where and how should Rust be used in Babel?

gaurav:
> ## ⚠️ AI-written experiment. Not reviewed by a human. Not a decision.
>
> Everything in this PR was written by Claude (Claude Code) at @gaurav's request. **No human has
> reviewed it yet**, including @gaurav, and **no decision has been taken**. Please read the
> recommendation as one input, not as an outcome.
>
> **The question it exists to answer:** #975 makes in-process pyo3 the way Rust enters Babel. Nothing
> had established that in-process is what Rust here actually needs. So: *where in Babel would Rust
> help, and how should it be wired in?*
>
> **The options being compared** (`docs/Rust.md` → "How to integrate"):
>
> | | Shape | Boundary |
> |---|---|---|
> | **A** | in-process pyo3, one package — **what #975 builds** | Python objects across FFI |
> | **B** | standalone program invoked from a Snakemake `shell:` rule | a file, at an existing rule edge |
> | **C** | split a rule at a *new* file boundary, implement one side in Rust | a file, at a new rule edge |
> | **D** | no Rust — DuckDB and other native libraries already in the tree | prebuilt wheel |
>
> **The recommendation is B/C by default, keeping A for the one case that needs it.** #975 currently
> implies A. That disagreement is the point of the PR and is for @gaurav and @SkyeAv to settle.
>
> This PR **does not touch #975's code** and does not ask for it to be reverted. It builds on top of
> it, keeps its plumbing, and only rewrites `docs/Rust.md` — a file an earlier AI pass wrote, whose
> two central factual claims turned out to be wrong (see below). That is itself a reason to check
> this document rather than trust it.

## What this adds

- **`docs/rust-decision/boundary_cost.py`** — the experiment, plus
  **`docs/rust-decision/README.md`** with its results and caveats.
- **`docs/Rust.md`** rewritten: where Rust would help, where it would not, the four shapes, the
  criteria that carry the decision, and the strongest argument *against* the recommendation.

No `src/` changes. No Rust code. Nothing in the pipeline behaves differently.

## The experiment, and why it does not benchmark pyo3

A pyo3 arm has no single honest shape — return a list of tuples? an iterator? a `#[pyclass]`? — so
any number invites *"you picked the wrong return type."* Instead the harness measures a **floor** no
pyo3 implementation can beat:

> **F** = the cost of constructing the Python objects the consumer needs, from raw bytes.

Anything that ends with Python holding the data pays at least F, so
**max possible saving from in-process = cost(file arm) − F**, with pyo3 credited a serialization cost
of exactly **zero**.

Payload: the anatomy clique state read straight out of a finished build's compendia (178,870 cliques
— a compendium line *is* a clique), replicated to multiply cardinality. Every arm ends holding
`set[frozenset[str]]`, what all 14 of `glom()`'s consumers build, and all arms were asserted to
produce an identical clique set before any timing was believed.

```text
=== scale 10x — 1,788,700 cliques, 3,069,240 members ===
arm         produce_s  consume_s   total_s   file_MB  peak_GB
floor            0.00       1.93      1.93       0.0     1.03
repr_set         1.30      10.50     11.80      53.4     1.01
jsonl            2.07       2.93      5.00      53.4     1.04
parquet          0.74       3.55      4.29      29.9     1.79
```

`repr_set` is not a straw man — `f"{set(s)}\n"` + `ast.literal_eval` is the *actual* format of the
boundary between `untyped_chemical_compendia` and `chemical_compendia` today.

## Result

Extrapolated to the chemical pipeline's **256,427,006 CURIEs** (`babel-1.18`'s
`reports/tables/cliques_table.csv`) against `chemical_compendia`'s measured **19,643 s**:

1. **The ceiling on what in-process could save is ~4 minutes — 1.3% of a 5.5-hour rule.** No pyo3
   implementation can do better, so boundary cost cannot justify in-process on its own.
2. **Switching the existing `repr(set)` boundary to JSONL would save ~9.5 minutes — 2.2× more than
   in-process could ever save**, with no Rust, no toolchain and no FFI. The serialization *format*
   matters more than the mechanism.

The floor also grows *faster* than the file arms across scales, so the relative case for in-process
shrinks at larger payloads rather than growing.

## Where Rust would help, and where it would not

- **Best case: `glom()`'s union-find.** The win is memory, not speed — glom holds clique state as
  Python objects at **73 MB resident for something that serialises to 4.8 MB (15×)**, which is why
  `protein_compendia` and `chemical_compendia` sit at 246 GB and 335 GB behind `mem="512G"`.
- **Not `generate_pubmed_concords`**, despite being the biggest number (20 h). It parses ~1,500
  *independent* files at one core — a sharding refactor, and @gaurav's DuckDB effort would supersede
  it anyway.
- **Profile `write_compendium` with `py-spy` before porting it.** Two uninvestigated defects sit in
  its setup: `get_biolink_model_toolkit` has no cache and refetches + reparses the Biolink YAML from
  GitHub on every call, and `InformationContentFactory` re-reads all 3,940,399 lines of the 212 MB
  `icRDF.tsv` calling `curies.compress()` per line — both **8 times** per chemical build.

## Two corrections to the earlier `docs/Rust.md`

Both were asserted by an earlier AI pass without checking, and both are load-bearing for the
conclusion that file was drawing:

- Rule counts are **243 `run:` and 26 `shell:`**, not "245 `run:` with no subprocess boundaries".
- **`reports.snakefile:209-213` already invokes a console script from `shell:`** with config-derived
  roots threaded in as `params:` — the exact pattern that paragraph claimed did not exist, and the
  precedent for how an out-of-process stage learns where its files are.

## What is not established

- **100× was not run.** It needs ~37M CURIEs; this ran on a 16 GB machine with 20 GB free. It should
  be repeated on a larger machine or an HPC node —
  `--scale 100` — and `docs/rust-decision/README.md` says what would falsify the write-up.
- The extrapolation is a per-item cost model, not a measurement at scale. It should be checked
  against a `py-spy` profile of a real run; if they disagree, believe the profile.
- I/O here is served from page cache, so it is not modelled; the real term is in the benchmark TSVs.
- **The one argument for in-process that survives these numbers** is the case where Rust keeps
  ownership and Python never materialises the data at all (a `#[pyclass]` iterated lazily). That
  escapes the floor entirely. Its out-of-process equivalent is "port the consumer too", so it is a
  question about how far a rewrite goes — and this experiment deliberately says nothing about it.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -266,7 +266,7 @@ ingest is in `docs/Development.md` ("Enhancing a data source ingest"); datahandl
   never export one that is called once per row, because crossing pyo3 per CURIE costs more than the
   Python it replaces. Every accelerated function keeps its Python implementation as the reference,
   with a test asserting the two agree, and is reached through `src/accel.py` (never `src._accel`
-  directly) so a missing extension falls back rather than breaking DAG parsing for all 245 rules.
+  directly) so a missing extension falls back rather than breaking DAG parsing for all 243 rules.
   Pick targets from a run's `benchmark:` TSVs, not by reading code — see
   [`docs/Rust.md`](docs/Rust.md), which also records the three targets chosen that way that turned
   out to be pure-Python bugs.
--- a/docs/Rust.md
+++ b/docs/Rust.md
@@ -1,7 +1,26 @@
 # Rust in Babel
 
-Babel's expensive rules are CPU-bound single-threaded Python. From the `babel-1.18` Snakemake
-`benchmark:` TSVs, the five costliest rules all run at ~100% of one core:
+> ## ⚠️ This document is an AI-written proposal. No human has reviewed or agreed to it.
+>
+> It was written by Claude (Claude Code) at @gaurav's request, to answer two questions that
+> [PR #975](https://github.com/NCATSTranslator/Babel/pull/975) leaves open: **where in Babel would
+> Rust actually help, and how should it be wired in?** It contains measurements, a recommendation,
+> and an argument against that recommendation.
+>
+> **Nothing here is a decision.** The options being compared are set out in
+> [How to integrate](#how-to-integrate) — in-process pyo3 (what #975 builds), a separate program at
+> a Snakemake rule boundary, splitting a rule at a new file boundary, or not using Rust at all. The
+> recommendation favours the second and third over the first; #975 currently implies the first.
+> Choosing between them is a call for @gaurav and @SkyeAv, who have not seen this yet.
+>
+> It also corrects two factual errors in an earlier AI-written version of this file, which is a
+> reason to check the rest rather than trust it. Every number below is reproducible from a committed
+> script or from `data/babel-1.18/babel_outputs/benchmarks/`.
+
+## Where the time actually goes
+
+From `babel-1.18`'s Snakemake `benchmark:` TSVs (354 rules, 83.7 h summed). All five costliest rules
+run at ~100% of a single core:
 
 | Rule | Wall | CPU | `mean_load` | `max_rss` |
 |------|------|-----|-------------|-----------|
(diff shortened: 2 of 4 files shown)


SkyeAv:
**Bottom line:** Taking most of this. But the headline — "in-process can save at most ~4 min /
1.3%, a ceiling no pyo3 can beat" — is a ceiling on one thing only: the boundary cost of a *single*
handoff. It says nothing about what a Rust implementation saves, because it assumes two things that
are false for how we use Rust.

**1. It converts one function, then reserializes.** It calls the cost of building `set[frozenset]`
in Python "an unavoidable floor." That floor is only incurred if you hand data back to Python
between operations. Chain the routines in Rust and the intermediate never materializes. Same two ops
on 1M cliques: round-trip-between-ops **12.4 s** vs fused-on-native-data **0.67 s** (−95%). The
"floor" is paid twice by the round-trip path, zero by the fused path. This is how Tablassert's Rust
works today (`build_fullmap_db` → `hydrate_*` → `lookup_*` share one native store).

**2. It's single-core.** The costliest rules run at ~100% of *one* core on many-core nodes. Same
Rust workload, only difference `py.allow_threads`: GIL held ~1.0× at every thread count; released,
near-linear to **10× at 16 threads**, one process. Tablassert ships this (a rayon pool behind pyo3).

**Accepted:** the boundary is cheap; the JSONL-over-`repr(set)` win (~9.5 min, no Rust); and Rule 0
(fix the Python first — pubmed `iterparse`, `conflate_synonyms` discarded-work, uncached
`get_biolink_model_toolkit`, 8× re-read of `icRDF.tsv`). But "the boundary is cheap" ≠ "in-process
saves only 1.3%."

**Ask:** Make A (in-process) the default mechanism, not a near-deprecated one — it is also the
cheaper thing to develop. One package, a function signature per routine that needs it, and a
byte-identical Python reference test; we write compiled code only for the steps that actually
benefit, and nothing else changes. B/C is the harder path: every accelerated stage becomes a new
file format to name, place, size into the disk budget and version, plus a new `shell:` rule and a
separate binary with its own build and SLURM reservation. The doc counts the build burden B/C spares
non-Rust contributors, but not the per-feature development cost B/C adds for whoever writes the
Rust. Reserve B/C for stages cleanly splittable at a boundary that already exists; use A everywhere
else. The doc's "one argument for in-process that survives" isn't a footnote — it's the main event,
and it's in production.

Fix: `lib.rs`/`accel.py` say "take down all 245 rules"; the real count is 26 `shell:` rules plus
~244 `run:`, so it should just read "every rule."

SkyeAv:
Closing without merging, per the analysis above. The experiment benchmarks a misuse of Rust — one function converted then immediately reserialized, single-threaded — so its "1.3% ceiling" bounds a single boundary crossing, not what a chained, GIL-released implementation saves (both patterns are already in production in Tablassert). It is also AI-written and unreviewed, and `docs/rust-decision/` plus a 313-line `Rust.md` rewrite would bloat the tree for a number we're rejecting.

The useful findings are carried as separate follow-ups: the JSONL-over-`repr(set)` boundary fix and the Rule 0 Python fixes (two of which are already open as #984 and #985). In-process pyo3 (#975) remains the default mechanism; #987's hardening of it is now merged.

PR #1043, opened 2026-08-19
Fix SAPBERT training data synonym duplication

### LINKED ISSUE #1011: Babel SapBERT training data contains many synonym pairs with identical terms
Example: ```biolink:CellLine||CLO:0022591||GM12814 cell||gm12814 cell||gm12814 cell``` in CellLine.txt

In fact, all synonym pair lines in CellLine.txt file contain identical terms, so they are all filtered out in my SapBERT training data creation script. Although my SapBERT training data creation script handles this without issues, the large amount of these identical terms in the produced Babel SapBERT training data blow up file sizes greatly. As Babel data grows larger with each new version, I think it'd be nice to not generate these identical terms in the first place to reduce file sizes. I am assigning it to myself to look into it when I get a chance, but @gaurav feel free to look into it as well if you get a chance before me.

hyi:
This PR should fix the identical synonym pair issue detailed in #1011 as well as deduplication of synonym pairs within the same biolink type. Closes #1011.

### DIFF
--- a/src/exporters/sapbert.py
+++ b/src/exporters/sapbert.py
@@ -62,6 +62,7 @@ def convert_synonyms_to_sapbert(synonym_filename_gz, sapbert_filename_gzipped):
     count_entry = 0
     count_training_rows = 0
     count_smaller_rows = 0
+    seen_pairs = set()
     with (
         gzip.open(synonym_filename_gz, "rt", encoding="utf-8") as synonymf,
         gzip.open(sapbert_filename_gzipped, "wt", encoding="utf-8") as sapbertf,
@@ -116,12 +117,17 @@ def convert_synonyms_to_sapbert(synonym_filename_gz, sapbert_filename_gzipped):
                 name_pairs = [(preferred_name_normalized, names[0])]                
             else:
                 name_pairs = list(itertools.combinations(set(names), 2))
+
+            name_pairs = [
+                name_pair for name_pair in name_pairs if (biolink_type, *sorted(name_pair)) not in seen_pairs
+            ]
             
             if len(name_pairs) > MAX_SYNONYM_PAIRS:
                 # Randomly select 50 pairs.
                 name_pairs = random.sample(name_pairs, MAX_SYNONYM_PAIRS)
 
             for name_pair in name_pairs:
+                seen_pairs.add((biolink_type, *sorted(name_pair)))
                 line = f"biolink:{biolink_type}||{curie}||{preferred_name}||{name_pair[0]}||{name_pair[1]}\n"
                 sapbertf.write(line)
                 count_training_rows += 1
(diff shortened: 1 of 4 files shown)


src/exporters/sapbert.py
```diff
@@ -62,6 +62,7 @@ def convert_synonyms_to_sapbert(synonym_filename_gz, sapbert_filename_gzipped):
     count_entry = 0
     count_training_rows = 0
     count_smaller_rows = 0
+    seen_pairs = set()
```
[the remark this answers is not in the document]
gaurav: Good catch — holding the name strings would have needed hundreds of gigabytes for GeneProteinConflated. Fixed in d95e2681: the set now holds a 64-bit digest per pair (`pair_key()`), which measures at ~95 bytes per entry against ~260+ for the tuple of strings, and the export rule declares `mem="64G"`. Kept it in memory rather than moving to a disk-backed unique index: the rule already takes 1.9h for this file, and several hundred million random-access inserts into a SQLite B-tree would likely cost more than the memory does. The `mem` figure is an estimate and should be rechecked against the benchmark TSVs after the first run that includes the deduplication.

gaurav (approved):
@hyi Looks good to me! I had Copilot and Claude go through it, and (1) they came up with the idea of storing hashes of the synonym pairs rather than the pairs themselves to save memory, (2) replacing the `random.sample()` I put in (which required enumerating every pair and then sampling 50) with a faster system that keeps sampling random pairs until it gets 50 unique pairs, which should be much faster, and (3) fixes some other bugs from my original code (e.g. I didn't check if `preferred_name` has pipe symbols in it). I'm approving the PR, but please have a look and see if you spot something that Claude got wrong before merging. Thanks!

src/exporters/sapbert.py
```diff
@@ -21,15 +21,96 @@
...
+def pair_key(biolink_type, name_pair):
+    """
+    Return a 64-bit digest identifying a (Biolink type, name, name) triple, independent of the
+    order of the two names.
+
+    We remember digests rather than the names themselves because the set of pairs already written
+    grows with the size of the output: GeneProteinConflated has hundreds of millions of cliques, so
+    holding on to the name strings would need hundreds of gigabytes, while the digests need roughly
+    a fifth of that. At 64 bits, the chance of even a single collision across a billion pairs is
+    around 3%, and a collision costs us one redundant training row.
```
hyi: I think the comment "a collision costs us one redundant training row" is confusing. It does not cost us one "redundant" training row, but potentially drops one otherwise valid training row since a distinct pair is mistakenly treated as an already seen pair due to the collision.

hyi (commented):
@gaurav Looks good to me. The memory concern for seen_pairs is definitely a valid one and keeping seen_pairs as hashes to reduce memory is a smart move.  I only left one very minor comment for your reference.

src/exporters/sapbert.py
gaurav: Well spotted! Fixed in f9afcf71 to "a collision costs us one correct training row". I'll go ahead and merge this now, but please feel free to open an issue or another PR if I screwed something up or if this needs further corrections.

Issue #1004, opened 2026-08-05
Simplify the three memory-bound "Factory"s


gaurav:
When downloading a Babel source, we often create additional files in the `babel_downloads/PREFIX/` directory:
- `labels`: labels for CURIEs that is loaded by [NodeFactory](https://github.com/NCATSTranslator/Babel/blob/b1b8a27de29365db0f284336c0a5a3657a1e7566/src/node.py#L446), but since the NodeFactory handles more complex tasks (like figuring out the Biolink Type for the CURIE), this probably shouldn't be fixed in this work.
- `synonyms`: synonyms for CURIEs, loaded by [SynonymFactory](https://github.com/NCATSTranslator/Babel/blob/b1b8a27de29365db0f284336c0a5a3657a1e7566/src/node.py#L26)
- `descriptions`: descriptions for CURIEs, loaded by [DescriptionFactory](https://github.com/NCATSTranslator/Babel/blob/b1b8a27de29365db0f284336c0a5a3657a1e7566/src/node.py#L120).
- `taxon`: which taxa a particular CURIE applies to (for genes, proteins, macromolecular complexes and human/mouse phenotypic features), loaded by [TaxonFactory](https://github.com/NCATSTranslator/Babel/blob/b1b8a27de29365db0f284336c0a5a3657a1e7566/src/node.py#L173).

Furthermore, we create `babel_downloads/common/[source]` files as well (currently, the only `source` we have is `ubergraph`): descriptions.jsonl, labels and synonyms.jsonl. This was created because a source like Ubergraph provides labels, synonyms and descriptions for many CURIEs with many prefixes. So the factories load all common descriptions/labels//synonyms, but a source-specific file will overwrite the common files.

These factories don't need a ton of memory for the smaller files (especially where they rely mostly on the common Ubergraph files rather than source-specific files), but this can be a strain for large files (e.g. taxa files for NCBIGene and ENSEMBL have taxa information for every gene and protein, which is 10s-100s of millions). I created a [TSVSQLiteLoader](https://github.com/NCATSTranslator/Babel/blob/b1b8a27de29365db0f284336c0a5a3657a1e7566/src/node.py#L190) that loads these TSV files using SQLite, allowing it to load large files without running out of memory. Only the TaxonFactory currently uses this.

This issue covers adding improved validation, disk storage and memory usage for these *Factory classes:
[...]

PR #1091, opened 2026-09-11
Add a Rust extension (maturin/pyo3) and port the SapBERT export to it, cutting its ~5.4 h of build time by more than half

gaurav:
Babel can now call Rust from a Snakemake rule. This PR takes the maturin/pyo3 plumbing from #975 (Skye's commits are carried over unchanged) and adds the one thing that PR's review asked for and never got: a real accelerator, wired in, provably equivalent to the Python it replaces. The accelerator is the SapBERT training-data export, `generate_sapbert_training_data_*`: 25 rules, about 5.4 h of summed cluster time per build in babel-1.18, all at ~99% CPU on one core, and an output we know is consumed (it trains RENCI-NER/sapbert).

Supersedes #975 and #588. Absorbs the conclusions of #988, which can be closed once this merges.

## Decisions worth reviewing

- **Why this function and not `glom()`, the `*Factory` loaders, or the encoding check.** Measured against the babel-1.18 benchmark TSVs and local files: `glom()` is the most important and least tested function in the pipeline and a wrong first port; returning a big dict from Rust cannot beat the cost of building that dict in Python (the #988 floor argument, now in `rust/README.md`), so the `*Factory` loaders (#1004) would gain nothing; and #953's Python encoding detector already runs at ~200 MB/s (all local labels and synonyms, 1.9 GB, in under 10 s), so Rust has nothing to add there. The SapBERT export is paths in, counts out, so nothing is materialised in Python at all, which is the one shape where in-process Rust escapes that floor. The KGX export was the runner-up (6.6 h) but its consumers are uncertain; babel-rdf is the natural second port.
- **Two deliberate differences from the Python, both marked `ponytail:` in `rust/src/sapbert.rs`.** Pair sampling is seeded per CURIE (FNV-1a of the CURIE), so a rerun is byte-identical; the Python's `random.sample` is unseeded and its `set(names)` order varies with `PYTHONHASHSEED`, so today's output is not reproducible even run-to-run. And gzip is written at level 6 rather than zlib's default 9, on flate2's `zlib-rs` backend; output size matched within 1%.
- **pyo3's `extension-module` feature is opt-in**, enabled only by maturin via `pyproject.toml`, so `cargo test` can link its test binary (it fails to link on Linux with the feature on). CI now runs `cargo test` next to fmt and clippy.
- **`ABI_VERSION` stays at 1.** The ABI has never shipped; bumping starts once it has.
- **The Python implementation stays in the tree for now** as the differential test's reference. Per `rust/README.md`, it is deleted once one full cluster build has run on the Rust path.

## What's here

- `rust/src/sapbert.rs`: the export in Rust, streaming gzip in (`MultiGzDecoder`, so concatenated members read the way Python's gzip does) and out, one `serde_json` parse per line, with unit tests for the name cleaning, pair counting, deterministic sampling and dedup.
- `src/exporters/sapbert.py` dispatches to Rust when the extension is built and to `_convert_synonyms_to_sapbert_python` otherwise (`BABEL_DISABLE_RUST=1` forces the fallback). It also drops the never-enabled `DrugChemicalSmaller` branch. **The Snakemake rule did not change**: it calls the same function it always did, and the function decides.
- `tests/exporters/test_sapbert.py` runs both implementations over five records copied verbatim from `Cell.txt.gz` plus synthetic records for branches the real files don't contain, and compares everything that is deterministic: leading fields, per-CURIE row counts, and the exact pair set for records with at most 50 pairs (the large majority). Sampled records are checked structurally. A `--pipeline` test does the same over every local `babel_outputs/synonyms/*.gz` and prints the A/B timings. `tests/test_accel.py` now asserts the built extension exports every accelerated function, so a forgotten `add_function` fails a test instead of silently falling back.
- `rust/README.md`: an accelerator table with the measured numbers, the SapBERT dispatch as the worked example, a section folding in #988's boundary-cost finding, and the history of why #975 closed. The #988 measurement script and write-up land under `docs/rust-decision/`, unchanged apart from links.

## What it produces

Measured on this laptop at 415cd4ba over `babel_outputs/synonyms/Disease.txt.gz` (354,071 entries, 1.3 M names); wall time equalled CPU time on every run:

| Implementation | Wall |
|---|---|
| Python, gzip level 9 (current) | 5.6 s |
| Python, gzip level 6 | 4.8 s |
| Rust, zlib-rs level 6 | 2.2 s |

The export is gzip-bound in both languages, so ~2.6× is close to the ceiling for this rule; on the cluster that is roughly 5.4 h becoming 2 h. The unchanged Snakemake rule was run end to end on `Cell.txt.gz`: the log shows `Using the Rust extension src/_accel (ABI_VERSION 1)`, the output is 49,834 five-field rows, and a second run reproduced it byte for byte. 445 unit tests pass; ruff, snakefmt, rumdl, cargo fmt, clippy and cargo test are clean.

## What it deliberately does not do

- It does not port `glom()`. Skye's Rust `glom()` (686 lines, on the #975 branch, run once on the cluster at `/projects/babel/runs/goetzs/RUST-GLOM-1.18`) was never evaluated; that is a follow-up, not something to fold in here.
- It does not touch the KGX export, babel-rdf, or the encoding check, and it does not delete the Python SapBERT implementation yet.

## Before merging

Nothing blocks this PR. The one condition on the *next* step is that the Python implementation is deleted only after a full cluster build has run the Rust path.

## Follow-ups (not yet filed as issues)

- Port [babel-rdf](https://github.com/frink-okn/babel-rdf) as the second accelerator and a first-class N-Triples output, using `oxrdf`/`oxttl` and the existing `curies` converter behind `get_biolink_prefix_map()` for the prefix map; decide with KGX's consumers whether it replaces `kgx/`.
- Port or drop the KGX export (6.6 h per build) depending on whether it keeps a consumer.
- Evaluate Skye's Rust `glom()` with `babel-clique-diff` against babel-1.18.
- Rebase #953 as Python-only and run its survey on a real build (locally it flagged 96 strings, 75 in `UMLS/synonyms`).
- The `chemical` / `protein` / `gene` rules are plain `gzip_files()` calls totalling ~4.4 h of single-threaded zlib; `pigz` or a lower compression level, no Rust.


docs/rust-decision/boundary_cost.py
SkyeAv, line 1: Not sure we need to include this in the final repo. I think it's bloat but that decision is down to you as future agents can use it.

releases/README.md
SkyeAv, line 1: What's this here? The date is in the past?

rust/src/sapbert.rs
SkyeAv, line 1: Claude might not have used the best decompression paradigm. It's stitching together the files it's unzipping and using a single thread to work on the stitched files. This is SUPER fast and cleaver for zipping the data but not unzipping it. I'm not sure how many dependencies you're okay with having but I'd point it at the gzp crate for what could be even faster compression, the rapidgzip-rust crate for better decompression, and the rayon crate for CPU bound synchronous parallelism.  By no means is what's here bad, these are just the kinds of optimizations you can do with a systems language like this.

src/exporters/sapbert.py
SkyeAv, line 1: I know this might be a massive shift but eventually we should probably use zstd compression as it's in python's standard lib while providing better and faster compression than gzip. This approach is fine as is though. It's a great use of passing data and avoids large serialization bottlenecks.

src/accel.py
SkyeAv, line 1: Not sure if this is intended but this implements the BABEL_DISABLE_RUST flag I thought we wanted to avoid? There is duplicate AB style code all throughout the PR which might functionally die out or drift, given enough use of the rust helpers.

SkyeAv (changes requested):

